PR #1, opened 2015-06-18
Clarity for API Description namespace

smizell:
[Rendered](https://github.com/refractproject/rfcs/blob/smizell/clarity-api-description/text/0000-clarity-api-description.md)

### DIFF
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -41,10 +41,7 @@ This type of transition, though, should have a URL and a method.
 
 We should add a section that defines what media types the server will respond with.
 
-# Unresolved questions
-
 ## Transition Attributes
 
-I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
-We have parameters, variables, and attributes throughout the document, which seems confusing.
-What should this be called?
+We should rename transition attributes to "data" to prevent confusion because the term "attributes" is already used within Refract.
+
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -8,65 +8,70 @@ Provide clarity and additional elements/properites in API Description namespace.
 
 # Motivation
 
-This will make it easier to understand for new users and cover more areas for existing API Description formats.
+The purpose of this document is to make it easier for newcomers to understand the format and build tools around it.
 
 # Detailed design
 
-I have spent a lot of time thinking about this namespace, and I believe I’m going to keep my suggestions for now mostly to making the namespace clearer to new users. I think we can handle adding state and affordances to this later to accommodate things like the concepts from Resource Blueprints.
-
-This contains mostly questions and thoughts. Would love your thoughts on what you'd like to see addressed in a PR.
-
 ## Combine Namespaces
 
-I think we should combine the API Description namespace with the resource namespace. Z is already good with this, but listing it here.
+We should combine the API Description namespace with the resource namespace.
+Right now they are separate and it would be clearer if they were not.
 
 ## Elements in Attributes
 
-This is mainly a point for discussion. Refract allows for including elements within the attributes section of an element. Even though it’s possible, would it be helpful to avoid this in such an important namespace? Reason being, this will be one of the first places people interact with Refract, and to use complex features may be too much.
+Refract allows for including elements within the attributes section of an element.
+Even though it’s possible, it may be helpful to avoid this in such an important namespace.
+Reason being, this will be one of the first places people interact with Refract, and to use complex features may be too much.
 
 ## HREF Variables
 
-I think we should call them “hrefVariables” everywhere instead of parameters in places. I think this will be more consistent and easier to understand for newcomers. This will make more sense in how we handle inheritance.
-
-Question: does it make sense to provide an “in” attribute that defines where in the URI template the variable may be found (e.g. path, query param, etc.)?
+We should call them “hrefVariables” everywhere instead of parameters in places (specifically transitions).
+I think this will be more consistent and easier to understand for newcomers.
+This will make more sense in how we handle inheritance as well, because parameters inheriting from hrefVariables does not seem optimal.
 
 ## Defining Classes
 
-Swagger provides a way to define tags that can be applied to resources throughout the document. Refract has the “class” attribute that we can use the same way, we only need a way to define a class name and provide the description for it. Of course, this won’t carry over to API Blueprint, but that’s OK.
+Swagger provides a way to define tags that can be applied to resources throughout the document.
+Refract has the “class” attribute that we can use the same way, we only need a way to define a class name and provide the description for it.
+Of course, this won’t carry over to API Blueprint, but that’s OK.
 
 ## Clarify Inheritance
 
-We need to clarify how inheritance works with:
+We need to clarify how inheritance works by defining the behavior for:
 
 * HREF Variables and HREF values throughout the document
 * Content types in payloads
 * Methods in transitions
 * Resources with schemas (though not in the spec) should be inherited by transitions
 
-I don’t so much mind the inheritance rules, as we can handle this kind of inheritance with Minim by traversing the tree up and down for the correct HREF.
-
 ## Transitions at Root
 
-We should be able to define transitions on the root of the document and not require they are nested within a resource. This type of transition, though, should have a URL and a method.
-
-## Transition Attributes
-
-I think it would be helpful to call this something other than attributes, as this will not be obvious to most people. Additionally, it means you have “element.attributes.attributes”, which is not a big deal, but something I think we can avoid. Not sure what to call it.
-
-This seems to me like a data structure for the transition.
+We should be able to define transitions on the root of the document and not require they are nested within a resource.
+This type of transition, though, should have a URL and a method.
 
 ## Subclass Asset
 
-I think it would be clearer if we subclass the Asset Element and added one called “messageShema” and another called “messageBody.” I think this will make reading a API Description make more sense than seeing Asset alone with classes.
-
-## Category Element
-
-I still have trouble understanding what “category” means here, but that may just be me. Is there a better name? Is this from the philosophical world? Maybe something like “section” or from the HTML world “div”?
+We should subclass the Asset Element and add one called “messageShema” and another called “messageBody.”
+This will make reading a API Description make more sense than seeing Asset alone with classes.
 
 ## Provided Media Types
 
-Swagger supports a section that defines what formats the server will respond with. This should be added (already discussed with Z).
+We should add a section that defines what media types the server will respond with.
 
 ## Add Schema to Resource
 
-Should resources be able to have their own schema (currently Asset with `messageBodySchema` as class).
+A resource should be able to have its own schema, whereas now it cannot.
+
+# Unresolved questions
+
+## Category Element
+
+What is a "category"?
+Does this word make sense used here?
+Maybe something like “section” or from the HTML world “div”?
+
+## Transition Attributes
+
+I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
+We have parameters, variables, and attributes throughout the document, which seems confusing.
+What should this be called?
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -1,0 +1,68 @@
+- Start Date: 2015-06-18
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+Provide clarity and additional elements/properites in API Description namespace.
+
+# Motivation
+
+This will make it easier to understand for new users and cover more areas for existing API Description formats.
+
+# Detailed design
+
+I have spent a lot of time thinking about this namespace, and I believe I’m going to keep my suggestions for now mostly to making the namespace clearer to new users. I think we can handle adding state and affordances to this later to accommodate things like the concepts from Resource Blueprints.
+
+This contains mostly questions and thoughts. Would love your thoughts on what you'd like to see addressed in a PR.
+
+## Combine Namespaces
+
+I think we should combine the API Description namespace with the resource namespace. Z is already good with this, but listing it here.
+
+## Elements in Attributes
+
+This is mainly a point for discussion. Refract allows for including elements within the attributes section of an element. Even though it’s possible, would it be helpful to avoid this in such an important namespace? Reason being, this will be one of the first places people interact with Refract, and to use complex features may be too much.
+
+## HREF Variables
+
+I think we should call them “hrefVariables” everywhere instead of parameters in places. I think this will be more consistent and easier to understand for newcomers. This will make more sense in how we handle inheritance.
+
+Question: does it make sense to provide an “in” attribute that defines where in the URI template the variable may be found (e.g. path, query param, etc.)?
+
+## Defining Classes
+
+Swagger provides a way to define tags that can be applied to resources throughout the document. Refract has the “class” attribute that we can use the same way, we only need a way to define a class name and provide the description for it. Of course, this won’t carry over to API Blueprint, but that’s OK.
+
+## Clarify Inheritance
+
+We need to clarify how inheritance works with:
+
+* HREF Variables and HREF values throughout the document
+* Content types in payloads
+* Methods in transitions
+* Resources with schemas (though not in the spec) should be inherited by transitions
+
+I don’t so much mind the inheritance rules, as we can handle this kind of inheritance with Minim by traversing the tree up and down for the correct HREF.
+
+## Transitions at Root
+
+We should be able to define transitions on the root of the document and not require they are nested within a resource. This type of transition, though, should have a URL and a method.
+
+## Transition Attributes
+
+I think it would be helpful to call this something other than attributes, as this will not be obvious to most people. Additionally, it means you have “element.attributes.attributes”, which is not a big deal, but something I think we can avoid. Not sure what to call it.
+
+This seems to me like a data structure for the transition.
+
+## Subclass Asset
+
+I think it would be clearer if we subclass the Asset Element and added one called “messageShema” and another called “messageBody.” I think this will make reading a API Description make more sense than seeing Asset alone with classes.
+
+## Category Element
+
+I still have trouble understanding what “category” means here, but that may just be me. Is there a better name? Is this from the philosophical world? Maybe something like “section” or from the HTML world “div”?
+
+## Provided Media Types
+
+Swagger supports a section that defines what formats the server will respond with. This should be added (already discussed with Z).
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -41,10 +41,6 @@ This type of transition, though, should have a URL and a method.
 
 We should add a section that defines what media types the server will respond with.
 
-## Add Schema to Resource
-
-A resource should be able to have its own schema, whereas now it cannot.
-
 # Unresolved questions
 
 ## Category Element
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -66,3 +66,7 @@ I still have trouble understanding what “category” means here, but that may 
 ## Provided Media Types
 
 Swagger supports a section that defines what formats the server will respond with. This should be added (already discussed with Z).
+
+## Add Schema to Resource
+
+Should resources be able to have their own schema (currently Asset with `messageBodySchema` as class).
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -43,12 +43,6 @@ We should add a section that defines what media types the server will respond wi
 
 # Unresolved questions
 
-## Category Element
-
-What is a "category"?
-Does this word make sense used here?
-Maybe something like “section” or from the HTML world “div”?
-
 ## Transition Attributes
 
 I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -49,11 +49,6 @@ We need to clarify how inheritance works by defining the behavior for:
 We should be able to define transitions on the root of the document and not require they are nested within a resource.
 This type of transition, though, should have a URL and a method.
 
-## Subclass Asset
-
-We should subclass the Asset Element and add one called “messageShema” and another called “messageBody.”
-This will make reading a API Description make more sense than seeing Asset alone with classes.
-
 ## Provided Media Types
 
 We should add a section that defines what media types the server will respond with.
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -23,12 +23,6 @@ We should call them “hrefVariables” everywhere instead of parameters in plac
 I think this will be more consistent and easier to understand for newcomers.
 This will make more sense in how we handle inheritance as well, because parameters inheriting from hrefVariables does not seem optimal.
 
-## Defining Classes
-
-Swagger provides a way to define tags that can be applied to resources throughout the document.
-Refract has the “class” attribute that we can use the same way, we only need a way to define a class name and provide the description for it.
-Of course, this won’t carry over to API Blueprint, but that’s OK.
-
 ## Clarify Inheritance
 
 We need to clarify how inheritance works by defining the behavior for:
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -17,12 +17,6 @@ The purpose of this document is to make it easier for newcomers to understand th
 We should combine the API Description namespace with the resource namespace.
 Right now they are separate and it would be clearer if they were not.
 
-## Elements in Attributes
-
-Refract allows for including elements within the attributes section of an element.
-Even though it’s possible, it may be helpful to avoid this in such an important namespace.
-Reason being, this will be one of the first places people interact with Refract, and to use complex features may be too much.
-
 ## HREF Variables
 
 We should call them “hrefVariables” everywhere instead of parameters in places (specifically transitions).


smizell:
@zdne I've updated the doc, FYI.

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+# Motivation
+
+The purpose of this document is to make it easier for newcomers to understand the format and build tools around it.
+
+# Detailed design
+
+## Combine Namespaces
+
+We should combine the API Description namespace with the resource namespace.
+Right now they are separate and it would be clearer if they were not.
```
zdne, line 18: :+1:

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Elements in Attributes
+
+Refract allows for including elements within the attributes section of an element.
+Even though it’s possible, it may be helpful to avoid this in such an important namespace.
+Reason being, this will be one of the first places people interact with Refract, and to use complex features may be too much.
```
zdne: :-1:

 I do not think this is possible. Especially when it comes to MSON samples / default values – those guys are and should be elements .

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+## HREF Variables
+
+We should call them “hrefVariables” everywhere instead of parameters in places (specifically transitions).
+I think this will be more consistent and easier to understand for newcomers.
+This will make more sense in how we handle inheritance as well, because parameters inheriting from hrefVariables does not seem optimal.
```
zdne, line 24: :ok_hand: 

I am neutral on this.

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Defining Classes
+
+Swagger provides a way to define tags that can be applied to resources throughout the document.
+Refract has the “class” attribute that we can use the same way, we only need a way to define a class name and provide the description for it.
+Of course, this won’t carry over to API Blueprint, but that’s OK.
```
zdne: :grey_question: 

Not sure what you are prospering here. Create an element that defines classes?

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+

+## Clarify Inheritance
```
zdne, line 26: :grey_question: 

I guess this is not really about inheritance (as in MSON). Are you talking about guidelines how to resolve or interpret some elements (hierarchy)? I do not understand any point bellow – maybe an very brief example would help?

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+We need to clarify how inheritance works by defining the behavior for:
+
+* HREF Variables and HREF values throughout the document
+* Content types in payloads
+* Methods in transitions
+* Resources with schemas (though not in the spec) should be inherited by transitions
+
+## Transitions at Root
+
+We should be able to define transitions on the root of the document and not require they are nested within a resource.
```
zdne, line 37: :+1:

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+This type of transition, though, should have a URL and a method.
+
+## Subclass Asset
+
+We should subclass the Asset Element and add one called “messageShema” and another called “messageBody.”
+This will make reading a API Description make more sense than seeing Asset alone with classes.
```
zdne: :ok_hand: 

I am somewhat neutral on this one. I can imagine other assets that are neither. But OK to proceed.

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Provided Media Types
+
+We should add a section that defines what media types the server will respond with.
```
zdne, line 42: :ok_hand: 

Isn't this duplicating the information? I guess I am failing to see a benefit of this. Is this the default media types for all requests / responses unless the request / response states otherwise?

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Add Schema to Resource
+
+A resource should be able to have its own schema, whereas now it cannot.
```
zdne: :-1: 

I am somewhat against this if it is just one schema. Because having a schema implies the serialization format. Another thing would be if a resource could have **multiple assets associated** and some of them could be schemas. That would be :+1:  for me.

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Unresolved questions
+
+## Category Element
+
+What is a "category"?
+Does this word make sense used here?
+Maybe something like “section” or from the HTML world “div”?
```
zdne: "Category" is a general grouping element. I am against "div" (HTML) or "group" and "section" (ambiguity with API Blueprint specification).

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Transition Attributes
+
+I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
+We have parameters, variables, and attributes throughout the document, which seems confusing.
+What should this be called?
```
zdne: So you want to get rid of parameters (in favor of hrefVariables). Given a transition is function its parameters are well, parameters. In representor we also use the term input properties.

zdne:
Done with the review. Please see my comments.

Legend: 
- :+1:  Yes. Let's do it.
- :-1:  No, I am against this.
- :ok_hand: I am easy on this one. 
- :grey_question: Need more clarification to understand.

text/0000-clarity-api-description.md
smizell: The URI Template RFC refers to them as "variables". I am more so interested in having consistency, and I'm still wondering why they are called parameters.

> Given a transition is function its parameters are well, parameters. In representor we also use the term input properties.

A URI template is instructions for how to build a request. It is more close to the concept of a form than a function, though when you implement you may _implement_ as a function. Either way, they are instructions.

---

But this point is saying that there is this other term used for transition attributes. This maybe should have been part of the unresolved questions area, but having parameters, variables, attributes, and data structures that are all simply MSON objects feels like we're missing something. I am not sure I have a proposal, but this seems like it could be quite confusing.

text/0000-clarity-api-description.md
smizell: To rephrase: what does "category" mean?

In the modern world, I see the term category used to provide additional meta data for some object, though I suppose traditionally it has meant as a way of grouping things. I'm fine leaving it, but it did take me a bit to understand what you meant by it.

text/0000-clarity-api-description.md
smizell: > Another thing would be if a resource could have multiple assets associated and some of them could be schemas

Sounds great.

Basically, it allows for using the [Resource Model](https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md#def-model-section). Swagger offers simply a way to define the body schema for an entire resource, so this would cover it as well.

text/0000-clarity-api-description.md
smizell, line 42: This is what the server MAY respond with. It is a list of hints at what the server provides, though the transactions may never show some of them.

Swagger has a property called [produces](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operation-object) that allows you to specify an array of available media types.

In reading over the Swagger spec again, I see that you can specify this `produces` on the API level. We do not have an API element to accommodate this, only `category`. This is a side question: should we have an `api` element for specifying this kind of API metadata?

text/0000-clarity-api-description.md
smizell, line 26: Yes, maybe better put as how to interpret certain values that are defined on parent elements. This is proposing we explicitly define those in the spec.

text/0000-clarity-api-description.md
smizell: Swagger has a [tags](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operation-object). Refract has classes. We can use classes, but have to have a way to define these classes used because they have meta data in Swagger.

The other option is to go the tag route and create tag elements, which is fine for me as well.

text/0000-clarity-api-description.md
smizell: Sounds good.

text/0000-clarity-api-description.md
smizell, line 24: This mainly goes to my point of consistency across things like parameters, attributes, and variables. Maybe we could agree on the best way to define these (which could be just to leave as is).

smizell:
@zdne I've address some comments. If you can check those out, let me know, and maybe we can move forward.

text/0000-clarity-api-description.md
zdne: @smizell Is this a general concept we would like to have in API Refract or is it more or less specific to a particular (swagger) format and as such it could be kept as part of [recently proposed metadata](https://github.com/refractproject/refract-spec/pull/30) perhaps using the `adapter` class?

text/0000-clarity-api-description.md
zdne, line 26: > how to interpret certain values that are defined on parent elements

:+1:

text/0000-clarity-api-description.md
zdne, line 42: :grey_question: 

Again, as above, the question is whether this is useful concept for any API (description / adapter) and thus it should be "canonical" or is it specific to Swagger format and hence it should be `adapter` metadata..

text/0000-clarity-api-description.md
zdne: :exclamation: I would like to deprecate and remove the Resource Model from API Blueprint completely so I would not build anything around this per se.

text/0000-clarity-api-description.md
zdne: Interestingly I've thought the first time I've heard the term was from you or Mark. Anyway, I am easy on settling on any other word if needed (but from those proposed I like category the best).

text/0000-clarity-api-description.md
zdne: > The URI Template RFC refers to them as "variables".

First, that RFC has many problems :roller_coaster: they call anything its value can vary in the template a "variable". However that does not say anything about what is the purpose – the use – of the variable. Is it a path segment or query parameter etc. Something can be hinted by operators. But I would definitely refrain from being inspired by RFC 6570. When it comes to names I would stick with URI RFC names instead. 

However without a concrete proposal I have not much to comment here.

zdne:
@smizell 

>  I've address some comments

I am not sure if you wanted to push any changes to this doc (if so a commit is missing?) or by "address" you simply mean you have add your comments on my comments. 

In any case, I've replied to your comments on my comments :laughing:

text/0000-clarity-api-description.md
pksunkara: I think calling them classes will be quite confusing.

text/0000-clarity-api-description.md
pksunkara: In my mind, I consider the MSON under the attributes section of a resource to be it's schema.

text/0000-clarity-api-description.md
pksunkara: :+1: I agree. Calling them attributes is quite confusing with refract's attributes. I generally say body params when I am explaining it to other users.

pksunkara:
I have reviewed and commented on with my opinions.

zdne:
@smizell what is the status of this ?

smizell:
@zdne I have unfortunately had no time to address this.

zdne:
@smizell OK but it is still open and we are totally doing it, right? :dancers:

smizell:
@zdne Yes! I want to do it, and think it needs to be done.

text/0000-clarity-api-description.md
danielgtaylor: Paraphrasing from @smizell & the [Swagger spec](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#tagObject): a resource may include a list of tags as a way of grouping items for tooling authors. These tags can contain a name, description, and external resource link. One resource may have multiple tags.

You can think of this as similar to, but more flexible, than our resource groups. A resource can only be in one resource group. Also, resource groups can create a hierarchy, while tags (as used in Swagger) cannot.

To me this is a generic API description feature. I believe that a flexible way to group resources would be beneficial to the adoption of a generic API description format.

I would propose that we either:
1. Create a hierarchical tag element for refract that references resources defined elsewhere in the document. Our resource groups would be implemented as this new element and we can support Swagger and other formats that use tags as well. For example:
   
   ``` yaml
   element: category
   meta:
   class: ['api']
   content:
   -
     element: category
     meta:
       class: ['resources']
     content:
       -
         element: resource
         meta:
           title: 'Resource 1'
         attributes:
           tags: ['tag1']
         content: ...
       -
         element: resource
         meta:
           title: 'Resource 2'
         attributes:
           tags: ['tag2']
         content: ...
   -
     element: category
     meta:
       class: ['tags']
     content:
       -
         element: tag
         meta:
           title: 'tag1'
           description: 'Some tag description'
         content:
           -
             element: tag
             meta:
               title: 'tag2'
               description: 'I am a nested tag'
           -
             element: externalRef
             content: 'http://example.com/'
   ```
   
   This is really powerful but could present some interesting problems, such as how the heck to render the documentation since a resource may be in multiple tags.
2. Leave things as-is and add a flat (i.e. non-hierarchical) tag element to support Swagger's use case. Tooling authors will need to support both resource groups and tags, but the tagging system is simplified. I think this is more explicit than re-using the `class` metadata. So far we have used the `class` metadata to differentiate the specific type of element (i.e. group of resources vs. group of data structures), not as an arbitrary grouping mechanism. **If** resources have a unique ID in the document then they should be easy to reference with a tag element.
   
   ``` yaml
   element: category
   meta:
   class: ['api']
   content:
   -
   element: category
   meta:
     class: ['resourceGroup']
   content:
     -
       element: resource
       meta:
         id: some-document-unique-identifier
       content: ...
   -
   element: category
   meta:
     class: ['tags']
   content:
     -
       element: tag
       meta:
         title: tag1
         description: I am a tag
         href: http://example.com/
       content: ['#some-document-unique-identifier']
   ```
   
   This will be simpler to implement but maybe less powerful. What do you think?

text/0000-clarity-api-description.md
danielgtaylor: Maybe we can also define when it is recommended to create a new element vs. using a class. This definition today seems to make sense:
- Create an element if it adds new metadata, attributes, or content.
- Otherwise, create a class if it adds nothing new but needs clarification on the type (e.g. generic grouping element `category` vs. a specific group of resources).

By that definition I do not believe that `messageSchema` or `messageBody` add anything so they should remain classes. I am open to changing the definition though :wink:

text/0000-clarity-api-description.md
danielgtaylor, line 42: > ...the question is whether this is useful concept for any API (description / adapter) and thus it should be "canonical" or is it specific to Swagger format

Another question: would something like this be valid and useful?

``` apib
# My resource [/frobs/{id}{?description}]
+ Attributes
  + id (required)
  + description
## Get a frob [GET]
+ Parameters
  + id (required)
+ Response 200 (application/json)
+ Response 200 (application/yaml)
+ Response 200 (application/xml)
+ Response 200 (text/plain)
## Update a frob [PUT]
+ Parameters
  + id (required)
  + description
+ Request ...
+ Response 200 (application/json)
+ Response 200 (application/yaml)
+ Response 200 (application/xml)
+ Response 200 (text/plain)
...
```

The basic question to ask here I think is whether the request/response examples are expected to be an exhaustive list of all possible returned mimetypes or if that information makes more sense at the root level of the API description document. I think it only makes sense at the root **if the responses can be generated**, e.g. from something like MSON with default values. Otherwise we have a documented & supported response mimetype with no usable example, and this is not specific to API Blueprint or Swagger.

text/0000-clarity-api-description.md
danielgtaylor: There are a few [usable choices](http://www.thesaurus.com/browse/group) but I'm not sure what was originally proposed. I'll agree that as an American English speaker the term `category` is a little awkward at first. I much prefer `group` or `collection`. Question: why does ambiguity with the API Blueprint spec pose an issue for Refract if the concepts are essentially the same?

text/0000-clarity-api-description.md
danielgtaylor: FWIW, this is also problematic for implementers that provide an `element.attributes` and want to expose shortcut properties. In [Fury this is a bit awkward](https://github.com/apiaryio/fury.js/blob/master/src/refract/api.es6#L258-L265):

``` js
// Long version
console.log(element.attributes.get('attributes'));

// Short version
console.log(element.attr);
```

If/when a Python or Ruby implementation happens we'll need to deal with the inevitable name collisions as well. If we could at least try to stay away from reusing these names that would be appreciated.

text/0000-clarity-api-description.md
```diff
@@ -0,0 +1,77 @@
...
+* Resources with schemas (though not in the spec) should be inherited by transitions
+
+## Transitions at Root
+
+We should be able to define transitions on the root of the document and not require they are nested within a resource.
+This type of transition, though, should have a URL and a method.
+
+## Subclass Asset
+
+We should subclass the Asset Element and add one called “messageShema” and another called “messageBody.”
```
kylef: Typo: `messageShema`

text/0000-clarity-api-description.md
kylef, line 37: Surely they are just transitions on the "root resource", i.e, with href of `/`?

text/0000-clarity-api-description.md
kylef: > @pksunkara I think calling them classes will be quite confusing.

I disagree with this point, they are called classes in HTML/CSS and it doesn't appear to have created that much confusion.

text/0000-clarity-api-description.md
kylef, line 42: I think this is somewhat linked to https://github.com/refractproject/rfcs/pull/1#discussion-diff-32811459R32, you could describe a class which states that various media types are supported.

text/0000-clarity-api-description.md
kylef, line 26: :+1: for this.

One note, tools such as swagger2blueprint should output when information is lost so the user knows what happened.

text/0000-clarity-api-description.md
smizell, line 37: I am not sure we can assume the URL (or resource) this transition belongs to. To me, it almost feels like shorthand for a resource with an HREF and a transition inheriting that HREF.

text/0000-clarity-api-description.md
smizell: @danielgtaylor Any thoughts on what would be good naming for it?

text/0000-clarity-api-description.md
smizell: Per @danielgtaylor's comment, we have:
- Category
- Group
- Collection

Which sounds best for this?

text/0000-clarity-api-description.md
smizell, line 42: This makes sense to me. Do you think we can address this by allowing for specifying media types as listed in this RFC? Do we need to do more?

text/0000-clarity-api-description.md
smizell: I'm good with using these definitions and leaving it as classes.

text/0000-clarity-api-description.md
smizell: @danielgtaylor The first to me seems less "magical," and it also seems more like a Refract way of doing it. But I see you've proposed nested tags. Is that the recommendation for doing this in a generic way?

smizell:
@danielgtaylor I've added a few comments throughout.

@kylef Relying on you to address these comments through some PR to the Refract namespaces. Thanks for doing that (in advance)!

text/0000-clarity-api-description.md
danielgtaylor: @smizell my thinking here is that if we want to support both generic hierarchical grouping elements (e.g. categories in refract today, which is how we implement API Blueprint resource groups) **and** allow an item to be in multiple grouping elements (a resource that is a part of two groups aka tags like in Swagger) then a way to support both is by creating a generic hierarchical grouping element that uses references to define its members. Rather than having an element like a resource group which contains resources, we have a tag element that a resource can reference. This allows for really flexible resource organization and should support most formats. For example:

``` apib
# Ungrouped Resource [/nogroup]

# Group Example Resources
Some description.

## My Resource [/foo]
## Another Resource [/bar]
```

Would become something like (even a bit simpler than I proposed above since we really don't need the `resources` or `tags` classes):

``` json
["category", {"class": ["api"]}, {}, [
  ["resource", {"title": "Ungrouped Resource"}, {}, ...],
  ["resource", {"title": "My Resource", "tags": ["Example Resources"]}, {"href": "/foo"}, ...],
  ["resource", {"title": "Another Resource", "tags": ["Example Resources"]}, {"href": "/bar"}, ...],
  ["tag", {"title": "Example Resources", "description": "Some description."}, {}, null]
]]
```

Of course that assumes we use the tag title as its unique identifier. As a tooling author I can build a menu based on the tags and using Minim we can easily query for all elements tagged with `foo`:
- Ungrouped Resource
- Example Resources
  - My Resource
  - Another Resource

Does that make sense? What do you think? This is somewhat future-proof as it allows arbitrarily nested levels of groupings and resources for organization while keeping the refract representation simple(-ish). It can also be extended in the future by adding a `class` to the `tag` element, for example to create a `dataStructure`-only tag or something along those lines. Any thoughts on this @zdne?

text/0000-clarity-api-description.md
danielgtaylor: @smizell for transition `attributes` I would suggest maybe `body`, `dataStructure`, or just `structure` or something along those lines. The term `attributes` is very API Blueprint-specific. What we are really describing is the data structure of the non-parameter (aka non-URL-based) input to the transition.

text/0000-clarity-api-description.md
pksunkara: I really like the idea in the above comment but I am worried about how documentation will render it if lots of resources from a resource group is mixed with data structures in between and resources from other resource groups. I guess we need to flesh out the example once to see if it's okay or not.

text/0000-clarity-api-description.md
pksunkara: I like `category`. If it's not possible, I would go with `collection`.

text/0000-clarity-api-description.md
pksunkara: We can use `body` if we use `query`. I don't think `dataStructure` or `structure` is a good idea.

kylef: I don't think we should use body, body is a HTTP term and this specification is generic and not HTTP specific.

From Roy's dissertation on REST Architectural Elements (5.2.1.2):

> REST components perform actions on a resource by using a representation to capture the current or intended state of that resource and transferring that representation between components. A representation is a sequence of bytes, plus representation metadata to describe those bytes. Other commonly used but less precise names for a representation include: document, file, and HTTP message entity, instance, or variant.
> 
> A representation consists of data, metadata describing the data, and, on occasion, metadata to describe the metadata (usually for the purpose of verifying message integrity). Metadata is in the form of name-value pairs, where the name corresponds to a standard that defines the value's structure and semantics. Response messages may include both representation metadata and resource metadata: information about the resource that is not specific to the supplied representation.

He uses the word "data" (structure?) to describe this, and I propose we do the same.

smizell: There are multiple contexts for this. I am good with `data` for describing the body of the transaction. What do you call URI template parameters? I think it would be good to sync these up throughout.

text/0000-clarity-api-description.md
smizell, line 26: Added as issue to Fury library.

https://github.com/apiaryio/fury.js/issues/39

text/0000-clarity-api-description.md
smizell: > Of course that assumes we use the tag title as its unique identifier.

I kind of struggle with this. On one hand, it makes sense because we would have to come with a unique ID for each tag. On the other hand, I get a funny feeling tying it to the human-readable title. I don't really feel strongly either way here.

I'm also a little iffy on having both category and tag in the spec. Seems like people have mental maps to what these are, and it would take some work to get your mind around the differences IMO. In other words, a resource can be in one and only one category, though it can be in many tags. (This is ignoring the fact the user could reference a resource with `ref`, so maybe I'm making a false assumption)

But I do really like this direction, and it seems like a necessity. We could leave this for a future change to the namespace so we can work this out more. @danielgtaylor @pksunkara

zdne: :grey_question: So is there any consensus on this? Leave `category` element for now and think about introducting "tags" in a future? Ideas @kylef ?

text/0000-clarity-api-description.md
zdne, line 37: I'd say this can be introduced later as a non-breaking change.

text/0000-clarity-api-description.md
zdne: For me to get pass this is either:
1. Do not introduce schemas for resources just yet (can be non-breaking later)
2. Introduce multiple schemas per resources (but do we really need it right now)

text/0000-clarity-api-description.md
zdne: All of the words are equally bad in my opinion so I do not have any clear winner. Anything is :ok_hand:  (collection has a different meaning when it comes to resources)

text/0000-clarity-api-description.md
zdne: Is this considering the Hypermedia Project charter and representor naming in any way?

zdne:
To me this PR is almost ready to go. The only really unresolved point seems to be the https://github.com/refractproject/rfcs/pull/1#discussion_r32812113 

The rest can be addressed later by introducing a non-breaking change. I would focus mainly on what we need now and make sure we won't have to break it tomorrow (e.g. any future changes we see happening will be non-breaking).

Thoughts?

text/0000-clarity-api-description.md
smizell: I believe we called them attributes when part of a resource. In regards to the URI template, looks like we bounce back and forth between attributes and parameters. 

I've also looked online in several places and do not see any consistency. [Here](https://msdn.microsoft.com/en-us/library/bb675245%28v=vs.110%29.aspx) they say things like query parameters and query string variables in the same article. I've seen this elsewhere. So I don't feel like we can go wrong.

So if we sync up with the Hypermedia Project, we use `attributes` for the body of a request/response. With regards to URI template, [Wikipedia](https://en.wikipedia.org/wiki/URL_Template) says parameters. Does that solve it? :)

smizell:
@zdne I agree. I think we should:
- Remove changes we'll address later
- Create other RFCs outlining those documents so we can discuss in further detail.

I think Kyle has already done some of the first.

text/0000-clarity-api-description.md
kylef: @smizell So the conclusion, we leave it as parameters and attributes?

smizell: @kylef Well, after thinking this weekend, I'm not entirely sure. In the Hypermedia Project, we call part of the payload attributes... not the entire payload. So `data` really may be better as it in encompasses attributes and anything else, such as transitions.

PR #6, opened 2015-08-03
Remove various items from Clarity API Description RFC

kylef:
This removes:
- Defining Classes
- Add Schema to Resource
- Category Element (proposed rename)

I believe this is what we're all agreeing to do, de-scope these items from the proposal and introduce later via separate RFCs in a backwards compatible manner.

### DIFF
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -41,10 +41,6 @@ This type of transition, though, should have a URL and a method.
 
 We should add a section that defines what media types the server will respond with.
 
-## Add Schema to Resource
-
-A resource should be able to have its own schema, whereas now it cannot.
-
 # Unresolved questions
 
 ## Category Element
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -43,12 +43,6 @@ We should add a section that defines what media types the server will respond wi
 
 # Unresolved questions
 
-## Category Element
-
-What is a "category"?
-Does this word make sense used here?
-Maybe something like “section” or from the HTML world “div”?
-
 ## Transition Attributes
 
 I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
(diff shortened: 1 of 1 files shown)


text/0000-clarity-api-description.md
```diff
@@ -47,18 +41,8 @@ This type of transition, though, should have a URL and a method.
...
 
-## Add Schema to Resource
-
-A resource should be able to have its own schema, whereas now it cannot.
-
 # Unresolved questions
 
-## Category Element
-
-What is a "category"?
```
pksunkara, line 58: Are we sure we want to remove this section? Don't we want to atleast answer the "What is category?"

kylef, line 58: This already seems clear in the [current specification](https://github.com/refractproject/refract-spec/blob/master/namespaces/api-namespace.md#category-element):

> ## Category (Element)
> 
> Grouping element – a set of elements forming a logical unit of an API such as
> group of related resources or data structures.
> 
> A category element MAY include additional classification of the category.
> The classification MAY hint what is the content or semantics of the category.
> The classification MAY be extended and MAY contain more than one classes.
> 
> For example a `category` element may be classified both as `resourceGroup` and
> `dataStructures` to denote it includes both resource and data structures.

/cc @smizell Is there any point I'm missing that's not documented correctly?

danielgtaylor, line 58: The description seems pretty clear to me. I believe the question was more around the naming.

smizell, line 58: Yes, my original section title was misleading I think. As Daniel said, it was mostly around the actual naming of the element. So if we're sticking with `category`, which it sounds like we are, this section may be removed.

text/0000-clarity-api-description.md
```diff
@@ -47,18 +41,8 @@ This type of transition, though, should have a URL and a method.
 
 We should add a section that defines what media types the server will respond with.
 
-## Add Schema to Resource
-
-A resource should be able to have its own schema, whereas now it cannot.
-
```
smizell, line 53: So are we never going to address this, or just not going to address it as part of this RFC? 

I believe Swagger [allows for this](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#path-item-object).

smizell, line 53: @kylef @pksunkara What about this?

pksunkara, line 53: I think we have decided not to address it as a part of this RFC.

smizell, line 53: Sounds good :)

PR #7, opened 2015-08-03
Rename transition attributes to data

kylef:
Just to clarify the conclusion on naming and move this point from the "Unresolved questions" section.

### DIFF
--- a/text/0000-clarity-api-description.md
+++ b/text/0000-clarity-api-description.md
@@ -41,10 +41,7 @@ This type of transition, though, should have a URL and a method.
 
 We should add a section that defines what media types the server will respond with.
 
-# Unresolved questions
-
 ## Transition Attributes
 
-I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
-We have parameters, variables, and attributes throughout the document, which seems confusing.
-What should this be called?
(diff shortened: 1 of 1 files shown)


text/0000-clarity-api-description.md
```diff
@@ -41,10 +41,7 @@ This type of transition, though, should have a URL and a method.
...
 We should add a section that defines what media types the server will respond with.
 
-# Unresolved questions
-
 ## Transition Attributes
 
-I think it would be helpful to call this something other than attributes, as this will not be obvious to most people.
-We have parameters, variables, and attributes throughout the document, which seems confusing.
-What should this be called?
+We should rename transition attributes to "data" to prevent confusion because the term "attributes" is already used within Refract.
```
smizell, line 46: This sounds good. One question I have is about URI template variables, which in the resource namespace are `hrefVariables`. Are we planning to keep that?

kylef, line 46: That's already address in the separate point (https://github.com/refractproject/rfcs/pull/1/files#diff-755b54829c9d98b897b8c1da6fb0b33aR20). Where we're going to rename them as `hrefVariables`.

smizell, line 46: Gotcha. So we'll have hrefVariables for URI templates and data for request/response bodies.

PR #8, opened 2015-08-06
Add RFC to rename class -> classes

danielgtaylor:
Based on the discussion in refractproject/refract-spec#37.

### DIFF
--- a/text/0000-class-rename.md
+++ b/text/0000-class-rename.md
@@ -1,0 +1,29 @@
+- Start Date: 2015-08-06
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes a backward-incompatible change of the `meta` property `class` to `classes`.
+
+# Motivation
+
+In many programming languages `class` is a reserved keyword. Sometimes it cannot be easily used as an attribute/property name (e.g. Python) and other times it clobbers useful existing information (e.g. Ruby). Changing to `classes` prevents both issues in most languages and also helps to explain that it is a list that can contain any number of class names.
+
+See also: https://github.com/refractproject/refract-spec/issues/37
+
+# Detailed design
+
+The meta property `class` will need to be renamed to `classes` in the base Refract specification. Additionally, the API Description and Parse Result namespaces will need to be updated to reflect this change. All documentation around the property needs to be updated. A new version of the Refract specification should be released, and its changelog should describe the breaking change.
+
+# Drawbacks
+
+This is a backward incompatible change and will require client libraries which implement the refract specification to be updated.
+
+# Alternatives
+
+Language-specific implementers can provide their own workarounds, for example you could provide a `cls` property in Python/Ruby instead of `class`. This may lead to fracturing and could become a maintenance burden over time.
+
+# Unresolved questions
+
+None.


kylef:
:+1:

text/0000-class-rename.md
```diff
@@ -0,0 +1,29 @@
+- Start Date: 2015-08-06
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes a backward-incompatible change of the name of the `meta` property `class` to `classes`.
```
pksunkara: `of the name of the` sounds weird.

zdne:
I am somewhat :ok_hand: with this. Not terribly happy about it but I am not against it 😬

text/0000-class-rename.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Motivation
+
+In many programming languages `class` is a reserved keyword. Sometimes it cannot be easily used as an attribute/property name (e.g. Python) and other times it clobbers useful existing information (e.g. Ruby). Changing to `classes` prevents both issues in most languages and also helps to explain that it is a list that can contain any number of class names.
+
+See also: https://github.com/refractproject/refract-spec/issues/37
+
+# Detailed design
+
+The meta property `class` will need to be renamed to `classes` in the base Refract specification. Additionally, the Resource, API, and Parse Result namespaces will need to be updated to reflect this change. All documentation around the property needs to be updated. A new version of the Refract specification should be released, and its changelog should describe the breaking change.
```
pksunkara: Resource and API namespaces are merged according to RFC 0002.

pksunkara:
Just some minor stuff. Otherwise lgtm.

text/0000-class-rename.md
pksunkara: And are now merged in refract-spec too.

danielgtaylor: Whoops. I'll update that.

danielgtaylor:
@pksunkara updated based on feedback.

smizell:
Personally, I would like to see these kinds of RFCs stay open just a little bit longer, as I actually didn't get around to reading it before it was merged (vacation and busy day). I am not sure @fosrias saw it either, which is who brought up the issue. Additionally, I'd like to hear why Z is unhappy with it, as in, is it because of the nature of breaking change? 

Lastly, I would want to add the question, can we foresee any other breaking changes? I definitely want to steer away from breaking changes as much as possible, but if we can anticipate any, we can make them now in one breaking version.

As in this this proposal, it works for me :)

In addition to this, I think it could be helpful to encourage people to implement this spec the best way possible in their language/platform. This may involve things like what Kyle showed me in Swift where he did not actually have a meta object on his element, but rather put each meta property directly on the class/struct. I don't see anything wrong with this, IMO. The shared understanding is in the messages, which is the representations of the state.

fosrias:
> This may involve things like what Kyle showed me in Swift where he did not actually have a meta object on his element, but rather put each meta property directly on the class/struct.

I  agree we need to not force a particular implementation. However, I do think (and this is not the first time I have seen this problem in an RFC) we should avoid using terms that collide with language keywords like the plague. E.g. `type` , `class` and the sort. These things run into problems.

PR #9, opened 2015-08-07
Cleanup files and rename RFC 0003

danielgtaylor:
Just a little cleanup and renaming the RFC 0003 file from 0000.


smizell:
Sorry, was swamped for a few days. Looks like we are moving forward with this change?

PR #10, opened 2015-08-11
Add RFC about API namespace clarification

pksunkara:
These are the issues and inconsistencies I found while working on API Blueprint to Refract conversion.

### DIFF
--- a/text/0000-clarify-api-namespace.md
+++ b/text/0000-clarify-api-namespace.md
@@ -3,57 +3,39 @@
 - Refract Issue: (leave this empty)
 
 # Summary
-Clarify API description namespace and Iron out issues which rose during API Blueprint serialization to API description namespace.
+Clarify API description namespace and iron out issues which rose during API Blueprint serialization to API description namespace.
 
 # Motivation
 The purpose of this document is to clear inconsistencies in the format and make it easier to build tools around it.
 
 # Detailed design
 
-## HREF Variables
-We need to clarify about the value of the `content` in the `hrefVariables` element.
-
 ## Data Structure
-We don't need a `Data Structure` because an element which is an `object`, `extend` or `array` can be considered as a data structure element.
+Correct the `Data Structure` in API description namespace to the following:
 
-*(The current spec is a lot confusing. What is the `Data Structure Element`? Why is `Data Structure` inheriting from it?)*
-
-## Resource
-Since `hrefVariables` is an element, we should move it to be under `content` rather than leave it under `attributes`.
-
-## Transition
-Similarily to the Resource, we need to move `hrefVariables` to be under `content`. `Data Structure` in Resource is under `content`, but it is under `attributes` in Transition. To resolve this inconsistency, we need to move `Data Structure` to be under `content`. Another reason for it is that `Data Structure` is an element.
+```apib
+# Data Structure (Element)
++ element: dataStructure (string, fixed)
++ content (Data Structure Element)
+```
 
 ## Transition Example
 Value of `content` should be changed to `[]` instead of `null`.
 
 ## Category
-We need to clarify what `transitions` category is about. If it is being reserved for future use, we need to add reserved keyword.
+Remove the "reserved" word from the description of `scenario` since it can be used as category semantics.
 
 ## Copy
-We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Is `description` just an aggregator of all `copy` elements? Why does some examples contain `description` without any `copy` elements and some examples contain `copy` elements but doesn't contain `description`. Since `copy` element is superior, why do we need `description` if we can use `copy` elements everywhere?
+The relation between copy elements in `content` and `description` in `attributes` should be clarified. Unless specified otherwise, a copy element's content represents the description of it's parent element and SHOULD be used instead of parent element's description metadata. `Copy Element` should be added to `content` of the following elements in API description namespace:
+
+- Category Element
+- Resource Element
+- Transition Element
+- HTTP Transaction Element
+- HTTP Message Payload Element
 
 ## HTTP Transaction Example
 Value of `content` should be changed to `[]` instead of `null`.
 
 ## HTTP Headers
-We need to change the base type from `Array Type` to `Element`. We also need to correct the `content` to be an `array[Member Element]`.
-
-## HTTP Message Payload
-Since `httpHeaders` is an element, we need to move it to be under `content`.
-
-# Drawbacks
-
-## Data Structure
-It can be tough for tools to get the data structure from the format.
-
-# Unresolved questions
-
-## Asset
-What is the `href` in `attributes`? What does it represent? Maybe we need to have an example there.
-
-## HTTP Transaction
-Why does it contain exactly 1 request and 1 response? A transaction example in API Blueprint can contain 1 request and many responses.
-
-## HTTP Request Message
-What about the name of a request from API Blueprint? How will it be represented?
+The `content` should be corrected to be an `array[Member Element]`.
--- a/text/0000-clarify-api-namespace.md
+++ b/text/0000-clarify-api-namespace.md
@@ -1,0 +1,59 @@
+- Start Date: 2015-08-10
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+Clarify API description namespace and Iron out issues which rose during API Blueprint serialization to API description namespace.
+
+# Motivation
+The purpose of this document is to clear inconsistencies in the format and make it easier to build tools around it.
+
+# Detailed design
+
+## HREF Variables
+We need to clarify about the value of the `content` in the `hrefVariables` element.
+
+## Data Structure
+We don't need a `Data Structure` because an element which is an `object`, `extend` or `array` can be considered as a data structure element.
+
+*(The current spec is a lot confusing. What is the `Data Structure Element`? Why is `Data Structure` inheriting from it?)*
+
+## Resource
+Since `hrefVariables` is an element, we should move it to be under `content` rather than leave it under `attributes`.
+
+## Transition
+Similarily to the Resource, we need to move `hrefVariables` to be under `content`. `Data Structure` in Resource is under `content`, but it is under `attributes` in Transition. To resolve this inconsistency, we need to move `Data Structure` to be under `content`. Another reason for it is that `Data Structure` is an element.
+
+## Transition Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## Category
+We need to clarify what `transitions` category is about. If it is being reserved for future use, we need to add reserved keyword.
+
+## Copy
+We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Is `description` just an aggregator of all `copy` elements? Why does some examples contain `description` without any `copy` elements and some examples contain `copy` elements but doesn't contain `description`. Since `copy` element is superior, why do we need `description` if we can use `copy` elements everywhere?
+
+## HTTP Transaction Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## HTTP Headers
+We need to change the base type from `Array Type` to `Element`. We also need to correct the `content` to be an `array[Member Element]`.
+
+## HTTP Message Payload
+Since `httpHeaders` is an element, we need to move it to be under `content`.
+
+# Drawbacks
+
+## Data Structure
+It can be tough for tools to get the data structure from the format.
+
+# Unresolved questions
+
+## Asset
+What is the `href` in `attributes`? What does it represent? Maybe we need to have an example there.
+
+## HTTP Transaction
+Why does it contain exactly 1 request and 1 response? A transaction example in API Blueprint can contain 1 request and many responses.
+
+## HTTP Request Message
+What about the name of a request from API Blueprint? How will it be represented?


zdne:
@pksunkara the RFC number should be `0004`. Also on many points I do not understand what the question and reasoning is. I am also missing any proposals on how to solve things. Can you please spent more time on explaining it? Thanks

pksunkara:
1. RFC numbers are added after they are merged.
2. I thought I added proposals for all the issues except the unresolved ones. Can you comment on them?

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+# Summary
+Clarify API description namespace and Iron out issues which rose during API Blueprint serialization to API description namespace.
+
+# Motivation
+The purpose of this document is to clear inconsistencies in the format and make it easier to build tools around it.
+
+# Detailed design
+
+## HREF Variables
+We need to clarify about the value of the `content` in the `hrefVariables` element.
```
pksunkara: There is nothing wrong with the `content` in `hrefVariables` element example. We just need write it's description.

zdne:
> RFC numbers are added after they are merged.

My bad, sorry

zdne:
> I thought I added proposals for all the issues except the unresolved ones. Can you comment on them? 

I am having hard times understanding the issues for example: 

> We need to clarify about the value of the content in the hrefVariables element.

Clarify what and where?

Furthermore you are mixing elements from different namespaces making the RFC really hard to follow – please try to think about how it is by somebody without the content. E.g. Data Structure Namespace Data Structure Element vs. API Namespace Data Structure etc.

If you are proposing a change please add reasoning, for example in:

> We need to change the base type from Array Type to Element 

I do not have any clue what is the reasoning here (and I would like to know).

pksunkara:
1. RE context: This RFC is all about API Namespace and it was described as such in the beginning.
2. I will add a bit more text for the reasoning.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+We need to clarify what `transitions` category is about. If it is being reserved for future use, we need to add reserved keyword.
+
+## Copy
+We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Is `description` just an aggregator of all `copy` elements? Why does some examples contain `description` without any `copy` elements and some examples contain `copy` elements but doesn't contain `description`. Since `copy` element is superior, why do we need `description` if we can use `copy` elements everywhere?
+
+## HTTP Transaction Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## HTTP Headers
+We need to change the base type from `Array Type` to `Element`. We also need to correct the `content` to be an `array[Member Element]`.
```
pksunkara: @zdne Well the reasoning is that `httpHeaders` is an element as shown in examples, but the description of it has `Array Type` which I thought was an accidental mistake. Are you saying that `httpHeaders` is actually an `Array Type`? Then why does the example have it as an element?

zdne:
@pksunkara also just because something is defined as an element it does not mean it cannot be used in un-refracted form (source map for example). 

Actually, given the current syntax, it is easier to define attributes as elements and then use them solely un-refracted.

text/0000-clarify-api-namespace.md
zdne: From API namespace: 

> The definition is a Data Structure namespace Object Type where keys are respective URI Template variables.

The type `content` of `hrefVariables` is the same as of `Object Type` – what is not clear about it?

pksunkara:
> Actually, given the current syntax, it is easier to define attributes as elements and then use them solely un-refracted.

But the examples show the respective elements in refracted form.

text/0000-clarify-api-namespace.md
pksunkara: Sorry, I got confused. `Array Type` is okay with me. But we really do need to change the description of `content`. Maybe make it similar to how `HREF Variables` was defined.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+## HREF Variables
+We need to clarify about the value of the `content` in the `hrefVariables` element.
+
+## Data Structure
+We don't need a `Data Structure` because an element which is an `object`, `extend` or `array` can be considered as a data structure element.
+
+*(The current spec is a lot confusing. What is the `Data Structure Element`? Why is `Data Structure` inheriting from it?)*
+
+## Resource
+Since `hrefVariables` is an element, we should move it to be under `content` rather than leave it under `attributes`.
```
zdne: :-1: No, it is an attribute to a resource's `href` not a content of a resource in any sense (in similar fashion you would have to move `href` there too). 

Just because something was defined as an element, it does not mean you cannot use it in attributes (ref our offline discussion about whether attributes may contain elements).

See 

> When attributes is an object, it MAY contain elements itself. The element definition SHOULD be used when interacting with attributes and its properites and values.

in https://github.com/refractproject/refract-spec/blob/master/refract-spec.md#base-element

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Detailed design
+
+## HREF Variables
+We need to clarify about the value of the `content` in the `hrefVariables` element.
+
+## Data Structure
+We don't need a `Data Structure` because an element which is an `object`, `extend` or `array` can be considered as a data structure element.
+
+*(The current spec is a lot confusing. What is the `Data Structure Element`? Why is `Data Structure` inheriting from it?)*
```
pksunkara: Don't consider this line. It slipped through in my commit.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Resource
+Since `hrefVariables` is an element, we should move it to be under `content` rather than leave it under `attributes`.
+
+## Transition
+Similarily to the Resource, we need to move `hrefVariables` to be under `content`. `Data Structure` in Resource is under `content`, but it is under `attributes` in Transition. To resolve this inconsistency, we need to move `Data Structure` to be under `content`. Another reason for it is that `Data Structure` is an element.
```
zdne: I feel there is a couple of things mixed here: 

:-1:  to "we need to move `hrefVariables` to be under `content` (see above)"

> `Data Structure` in Resource is under `content`, but it is under `attributes` in Transition.

There is fundamental difference between the two – data structure in Resource describes the content of a resource. While in transition the data structure describes input parameters (attributes) of the transition. 

:-1: to "To resolve this inconsistency, we need to move `Data Structure` to be under `content`."

and as mentioned above "Another reason for it is that `Data Structure` is an element." is not a reason for thing to be in content.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Transition Example
+Value of `content` should be changed to `[]` instead of `null`.
```
zdne, line 23: :ok_hand:

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Category
+We need to clarify what `transitions` category is about. If it is being reserved for future use, we need to add reserved keyword.
```
zdne: I think this was introduced by @kylef or @smizell in the call for being able to define transitions at the root level. I am up for any clarification if needed, but frankly I think it is not. It is just a group of transitions. It does not have a syntax in API Blueprint and neither scenarios have. 

I am :+1: to **REMOVE** the "reserved" word from "scenario - Reserved. Category is set of steps." since it is not really reserved. It can be used as a category semantics. Just because something does not have a counter part in API Blueprint syntax it does not mean it has to be reserved here. (Simply the conversion will ignore it)

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+# Motivation
+The purpose of this document is to clear inconsistencies in the format and make it easier to build tools around it.
+
+# Detailed design
+
+## HREF Variables
+We need to clarify about the value of the `content` in the `hrefVariables` element.
+
+## Data Structure
+We don't need a `Data Structure` because an element which is an `object`, `extend` or `array` can be considered as a data structure element.
```
zdne: Tricky one, let me get to it at the end.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+Value of `content` should be changed to `[]` instead of `null`.
+
+## Category
+We need to clarify what `transitions` category is about. If it is being reserved for future use, we need to add reserved keyword.
+
+## Copy
+We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Is `description` just an aggregator of all `copy` elements? Why does some examples contain `description` without any `copy` elements and some examples contain `copy` elements but doesn't contain `description`. Since `copy` element is superior, why do we need `description` if we can use `copy` elements everywhere?
+
+## HTTP Transaction Example
+Value of `content` should be changed to `[]` instead of `null`.
```
zdne, line 38: :ok_hand:

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+Similarily to the Resource, we need to move `hrefVariables` to be under `content`. `Data Structure` in Resource is under `content`, but it is under `attributes` in Transition. To resolve this inconsistency, we need to move `Data Structure` to be under `content`. Another reason for it is that `Data Structure` is an element.
+
+## Transition Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## Category
+We need to clarify what `transitions` category is about. If it is being reserved for future use, we need to add reserved keyword.
+
+## Copy
+We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Is `description` just an aggregator of all `copy` elements? Why does some examples contain `description` without any `copy` elements and some examples contain `copy` elements but doesn't contain `description`. Since `copy` element is superior, why do we need `description` if we can use `copy` elements everywhere?
```
zdne: :+1:

text/0000-clarify-api-namespace.md
zdne: :grey_question: Why?

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+## HTTP Transaction Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## HTTP Headers
+We need to change the base type from `Array Type` to `Element`. We also need to correct the `content` to be an `array[Member Element]`.
+
+## HTTP Message Payload
+Since `httpHeaders` is an element, we need to move it to be under `content`.
```
zdne: :-1: no, see above

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Drawbacks
+
+## Data Structure
+It can be tough for tools to get the data structure from the format.
```
zdne: :grey_question: Not sure what do you mean here.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Unresolved questions
+
+## Asset
+What is the `href` in `attributes`? What does it represent? Maybe we need to have an example there.
```
zdne: If an asset is remote, this will be its URL.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+## HTTP Transaction
+Why does it contain exactly 1 request and 1 response? A transaction example in API Blueprint can contain 1 request and many responses.
```
zdne: This has been long called for by testing. Basically it disambiguate things for tooling authors. For conversion it means things has to be copied (duplicated) :(

zdne: If we want to challenge this I'd open the discussion with @ntmlk

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,59 @@
...
+
+## HTTP Request Message
+What about the name of a request from API Blueprint? How will it be represented?
```
zdne: In the `name` property?

text/0000-clarify-api-namespace.md
pksunkara: > > We need to change the `content` description in HTTP Headers
> 
> Why?

Because, it is currently wrong. Check the examples and then check the description.

text/0000-clarify-api-namespace.md
pksunkara: I mean that, if we go along with my proposal, the tools which work on this format need to check for many conditions to recognise a data structure.

text/0000-clarify-api-namespace.md
pksunkara: I still don't understand this.

text/0000-clarify-api-namespace.md
pksunkara: I am okay with duplicating. :heavy_check_mark:

text/0000-clarify-api-namespace.md
pksunkara: Sorry. Forgot about that. :disappointed:  :heavy_check_mark:

text/0000-clarify-api-namespace.md
pksunkara: Got it. Just got confused. :heavy_check_mark:

text/0000-clarify-api-namespace.md
smizell, line 23: This is just a note, no action needed.

We have talked before about having a best practices for serializing/parsing JSON representations of Refract. If so, we should define how we interpret `null` in these situation.

pksunkara, line 23: > This is just a note, no action needed.

So, we just leave it as `null`? I don't like that. I think it should be empty array or the key `content` should not be defined.

text/0000-clarify-api-namespace.md
smizell: When you say `name` property, to which are you referring?

pksunkara: The `name` of a request.

```apib
+ Request New User (application/json)

        {}
``

File: text/0000-clarify-api-namespace.md (hunk 14 above)
Comment (smizell) [inline text/0000-clarify-api-namespace.md]: Just a question: can you safely copy requests or responses to additional transactions and retain original information? Does API Blueprint define any guidelines for how multiple requests/responses should be handled?

File: text/0000-clarify-api-namespace.md (hunk 15 above)
Comment (pksunkara) [inline text/0000-clarify-api-namespace.md]: Good point. We are using `attributes.title` for resource and action names. Not a `name` property.

File: text/0000-clarify-api-namespace.md (hunk 6 above)
Comment (smizell) [inline text/0000-clarify-api-namespace.md:23]: Sorry for the confusion here, was trying to say, "I just want to add a note to this, you don't have to do anything with my note."

I'm good with changing it to empty array.

File: text/0000-clarify-api-namespace.md (hunk 14 above)
Comment (pksunkara) [inline text/0000-clarify-api-namespace.md]: I think we should be able to safely copy the requests and responses. But the main issue here is since we allow m-to-n request response pairs, we will end up higher duplications.

``` apib
+ Request A

+ Request B

+ Response 201

+ Response 200
```

The above snippet will give out 4 http transactions :(

text/0000-clarify-api-namespace.md
zdne: > cause an element which is an `object`, `extend` or `array`

Note it can be any time (even `string` etc).

You are correct that as it is defined here it does not make much sense. This element was supposed to represent "Attributes" section in API Blueprint. Correctly it should be:

``` apib
# Data Structure (Element)
+ element: dataStructure (string, fixed)
+ content (Data Structure Element) 
```

so, for example: 

``` json
["dataStructure", {}, {}, [
  "string", {}, {}, "Hello World!"
  ]]
```

The reason to have a separate `Data Structure` element instead of just directly a _Data Structure Namespace_ `Data Structure Element` was to explicitly state "this is data structure defined to represent the (parent) element" versus just having an arbitrary `object` or `string` lying there:

Consider:

``` json
["resource", {}, {}, [
  ["dataStructure", {}, {}, [
    "string", {}, {}, "Hello World!"
    ]]
  ]]
```

vs.

``` json
["resource", {}, {}, [
  "string", {}, {}, "Hello World!"
  ]]
```

Thoughts?

pksunkara: I am okay with the separate `Data Structure` element.

text/0000-clarify-api-namespace.md
zdne: I suggest following: 

> # Copy (Element)
> 
> Copy element represents a copy text – a textual information content of an element. Copy element MAY include information about the media type of the copy's content. 
> 
> Unless specified otherwise, a copy element's content represents the description of it's parent element and SHOULD be used instead of parent element's description metadata.

In addition to this I suggest to explicitly add Copy as a possible value for `content` arrays of:
- Category element
- Resource element
- Transition element
- HTTP Transaction element
- HTTP Message Payload element

---

However, if you are following, it is clear this does not solve MSON – Data Structure Namespace and HTTP headers. The reason for this is that currently it is not possible to add Copy elements to primitive types. 

Consider theoretical future syntax (`description` keyword): 

``` apib
# Person (object)
## Description
Copy1 

## Properties
- name

## Description
Copy 2
```

With current refract namespace this will cause us a lot of headache. Not necessary a problem of the `copy` element discussed here, but something we need to consider.

text/0000-clarify-api-namespace.md
zdne: Almost feel like we should actually revisit the Data Structure namespace but that is "oh my" moment. See my comment on copy https://github.com/refractproject/rfcs/pull/10/files#r36758614

text/0000-clarify-api-namespace.md
pksunkara: Works for me. :smile:

smizell: Is the question then, how should we handle these elements where they seem to be content but are not allowed as content by the other namespaces?

text/0000-clarify-api-namespace.md
zdne: sure it is the `attributes.title` 

all the renaming! :angry:

text/0000-clarify-api-namespace.md
zdne: an asset does not have to be a local thing. It can be a JSON somewhere in the internet. So you won't have body just its URL.

text/0000-clarify-api-namespace.md
smizell: So is the guideline that this below means four transactions:

``` apib
+ Request A
+ Request B
+ Response 201
+ Response 200
```

and this below means two?

``` apib
+ Request A
+ Response 201
+ Request B
+ Response 200
```

pksunkara: Yup.

text/0000-clarify-api-namespace.md
smizell: I too vote for data structure because it actually conveys more information than simply being primitive values. This is because, a data structure can have different inheritance rules than normal Refract structure. This is OK, and the element creates a clean separation of that IMO.

text/0000-clarify-api-namespace.md
smizell: @zdne is correct here. Element can be found in attributes or content. The purpose is to allow for annotating even attributes with their own attributes.

text/0000-clarify-api-namespace.md
smizell: Two things:
1. It is actually not wrong to have it as an array like this, IMO. An Array Element is simply an element with it's content as an array.
2. In the [current namespace](https://github.com/refractproject/refract-spec/blob/master/namespaces/api-description-namespace.md#http-headers-array-type), the `Properties` section is incorrect. As @pksunkara stated, it should be `array[Member Element]`.

smizell: One more thing, the HTTP Message Payload is defined extending upon Element, so it may be good to sync these up. I would say they are both arrays, because that is easier to implement.

It's easier because HTTP Message Payload can just extend the current Array Element rather than having to define every method as if it was its own primitive. (if that makes sense)

text/0000-clarify-api-namespace.md
zdne: >  how should we handle these elements where they seem to be content but are not allowed as content by the other namespaces?

What you refer to by "these" elements?

zdne: Let me rephrase the problem: I do not know how to add a copy element to string data structure element

text/0000-clarify-api-namespace.md
pksunkara: But HTTP Message Payload's content is not `array[Member Element]`. So, I think leaving its base type as `Element` is the correct thing.

smizell: It's OK that isn't `array[Member Element`, because that is usually for Object Elements.

For me, the correct element for HTTP Message Payload to inherit from is `Array Type` and the content of it should be an `array`. The HTTP Headers element should inherit from `Array Type` as well and have `array[Member Element]` as its content.

pksunkara: Well, the content of Copy or Category elements is also an array, so why are we not making their base type `Array Type`? I am having a tough time understanding why you are suggesting that only a select few elements inherit `Array Type`.

text/0000-clarify-api-namespace.md
pksunkara: @zdne I think Stephen's question is about the "Unless specified otherwise". We haven't specified anything currently in any other elements. So, that does that mean all of them should use copy element? What about the elements in the data structure namespace?

pksunkara:
I have addressed the reviews and incorporated the changes into a commit. I haven't pushed the commit since I am still waiting on this [comment](https://github.com/refractproject/rfcs/pull/10#discussion_r36775893)

text/0000-clarify-api-namespace.md
smizell: @pksunkara Good question! In my opinion, elements should always inherit from a primitive element. This is important to me for implementation, as you know a certain element can extend a primitive class. Otherwise, you have to implement all methods for that new element OR make an assumption that it is of a certain primitive type.

pksunkara: Yes. So, why are we making `HTTP Message Payload` inherit `Array Type`? I feel like I am missing something in the communication.

pksunkara: `Category` can inherit from `Array Type` too, isn't it?

smizell: Well, for one, I think all elements should somehow inherit from a primitive. This seems right to me, but open for discussion for scenarios where this isn't possible.

Additionally, if you inherit from a primitive, it means your code will look very similar.

See https://github.com/apiaryio/fury.js/blob/master/src/refract/api.es6#L305

So even though the spec doesn't say, it is clear this is how this turns out because you want all of the same properties (methods) that an array type has.

pksunkara: To be honest, in the C++ drafter code, we are creating Category element out of `Array Type`. I think we should do another RFC after this is settled.

smizell: @pksunkara Sounds good to me. Maybe one on making sure we are inheriting from the correct elements?

pksunkara: Yup.

smizell: :+1:

text/0000-clarify-api-namespace.md
smizell: I'm wondering if it may be beneficial to have a video chat about this, because there is actually a lot going on in this.

My first question is, why would you add a copy element to a string? To me, a string is a string. The difficulty here is that we have diverged some from the original structure of the MSON document, though staying with the structure is equally as difficult. Take for instance your example.

``` md
# Person (object)
## Description
Copy1 

## Properties
- name

## Description
Copy 2
```

If you didn't have these descriptions, everything would be fine. But once you add them, you also want to retain their structure. This translates to:
- Structure Declaration
  - Copy (Description)
  - Properties
  - Copy (Description)

If you want to retain this order, I'm not sure you can get around this without mimicking this structure. But this really complicates things.

``` js
["category", { "class": "element" }, {}, [
  ["copy", {}, {}, "Copy1"],
  ["object", { "id": "Person" }, {}, [
    ["member", {}, {} {
      "key": ["string", {}, {}, "name"],
    }]
  ]],
  ["copy", {}, {}, "Copy2"]
]]
```

Not really proposing this, but my gut feeling is that if you want a solution, you're going to have to go this route or come up with something even more complex.

text/0000-clarify-api-namespace.md
zdne: @smizell @pksunkara let me clarify few things: 

> I feel like I am missing something in the communication.

The missing piece is "MSON all the thing" – my vision is that in API Blueprint, Attributes, Parameters and Headers are nothing else but just Data Structure definitions. 

Let's look at it in detail: 

## Parameters

The top level data structure representing parameters is a hash of parameters – hence `Href Variables` is of  `Object Type` type.

## Attributes

The top level data structure representing anything can be any type – hence `Data Structure` was of `Data Structure Element` type. 

## Headers

The top-level data structure representing headers is an array of key-value pairs – hence `HTTP Headers` is of `Array Type` type.

The situation gets complex for headers because unlike with parameters you can have multiple HTTP headers of the same name. See the open question in my following comment.

# Background

Imagine headers, attributes and parameters are interconnected – the hypothetical blueprint:

``` apib

# Data Structures
## Page Size (number)
Number of items per a page.

# Group Resources
## Collection [/collection]
+ parameters
    + page_size (Page Size)

### Retrive [GET]
+ response 200
    + headers
        + page_size (Page Size)
    + attributes 
        + items (array)
        + page_size (Page Size)
```

zdne: :grey_question: Question apparently the current spec mixes two approaches:
1. Header name is defined as title in meta attributes

``` json
["httpHeaders", {}, {} [
  ["string", {"title": "Content-Type"}, {}, "application/json"]
  ]]
```
1. Header name is defined as key using the Member Element

``` json
["httpHeaders", {}, {} [
    ["member", {}, {} {
      "key": ["string", {}, {}, "Content-Type"],
      "value": ["string", {}, {}, "application/json"]
    }]
  ]]
```

Not sure which one is better, given the header value should be any data structure element. Ideas?

pksunkara: Given how we are formatting `hrefVariables`, I would go similarly with `httpHeaders`. But instead of an `Object Type`, it's an `Array Type`.

text/0000-clarify-api-namespace.md
zdne: I am not saying we have to go this way. I am just raising concerns that if we stick with my proposal https://github.com/refractproject/rfcs/pull/10#discussion_r36758614 then in a future we won't have a backward compatible way to introduce multiple copy elements to some elements in _Data Structure Namespace_.

text/0000-clarify-api-namespace.md
zdne: > But the main issue here is since we allow m-to-n request response pairs, we will end up higher duplications.

That is the consequence, yes.

pksunkara:
Addressed the reviews and updated.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,35 @@
+- Start Date: 2015-08-10
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+Clarify API description namespace and Iron out issues which rose during API Blueprint serialization to API description namespace.
```
zdne: nit: Iron -> iron

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+## Transition Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## Category
+Remove the "reserved" word from the description of `scenario` since it can be used as category semantics.
+
+## Copy
+We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Unless specified otherwise, a copy element's content should represent the description of it's parent element and SHOULD be used instead of parent element's description metadata. And a `Copy Element` should be added to `content` of other elements of API namespace wherever possible.
```
zdne: Remove "And a" in "And a `Copy Element`"

Change "wherever possible" to the specific element discussed in the PR

zdne:
Two comments otherwise looks good to me.

pksunkara:
Updated.

pksunkara:
Please merge refractproject/refract-spec#41 once this is merged.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,41 @@
+- Start Date: 2015-08-10
+- RFC PR: (leave this empty)
```
zdne, line 2: I guess you can add a link to PR since you have it.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,41 @@
...
+# Summary
+Clarify API description namespace and iron out issues which rose during API Blueprint serialization to API description namespace.
+
+# Motivation
+The purpose of this document is to clear inconsistencies in the format and make it easier to build tools around it.
+
+# Detailed design
+
+## Data Structure
+Since `Data Structure` inheriting from `Data Structure Element` of DataStructure Correct the `Data Structure` to the following:
```
zdne: This sentence does not make much sense as such. Also I do not think "DataStructure" is used anywhere else.

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Transition Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## Category
+Remove the "reserved" word from the description of `scenario` since it can be used as category semantics.
+
+## Copy
+We need to clarify what the relation between copy elements in `content` and `description` in `attributes` is. Unless specified otherwise, a copy element's content should represent the description of it's parent element and SHOULD be used instead of parent element's description metadata. `Copy Element` should be added to `content` of the following elements in API namespace:
```
zdne: "should represent" :arrow_right: "represents" 
"API namespace" :arrow_right: "API Description namespace"  for consistency...

text/0000-clarify-api-namespace.md
```diff
@@ -0,0 +1,41 @@
...
+- Resource Element
+- Transition Element
+- HTTP Transaction Element
+- HTTP Message Payload Element
+
+## HTTP Transaction Example
+Value of `content` should be changed to `[]` instead of `null`.
+
+## HTTP Headers
+We need to correct the `content` to be an `array[Member Element]`.
```
zdne: Stylistics: 

Seems you have two "voices" in this document:
1. X should be changed to Y
2. We need X to be Y 

While I get what you are trying to say it feels like one is a recommendation while the other is a must.

zdne:
@pksunkara almost there! Main objection please re-read and fix the sentence on Data Structure

text/0000-clarify-api-namespace.md
pksunkara, line 2: This and RFC number update is done by Stephen only after the PR is merged. So, I will leave it.

pksunkara:
Updated.

smizell:
@pksunkara Is there anything left to address here? Looks like you've addressed @zdne's comments. If so, let's merge so implementations can be merged.

zdne:
@smizell looks good. Merging.

PR #13, opened 2015-08-29
RFC for steps to move to Refract 1.0

smizell:
This document outlines some things we should consider as we move Refract to production environments.

### DIFF
--- a/text/0000-refract-1.0.md
+++ b/text/0000-refract-1.0.md
@@ -1,0 +1,56 @@
+- Start Date: 2015-08-29
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+The document provides some thoughts on how we can move forward to version 1.0 of Refract. This document is not proposing we move to 1.0 immediately, but to rather start putting steps in place to get us there quickly and being intentional about it.
+
+# Motivation
+
+As Refract is moving into production environments, we need to keep the following in mind:
+
+1. We want to have as a concise of a spec as necessary for newcomers
+1. We want to introduce breaking changes thoughtfully, which is what SemVer encourages. We currently can't do SemVer correctly without moving to 1.0.
+1. We want to start registering media types for Refract representations
+1. Moving this to production means people are actually going to start using this spec in the wild, and we need to prepare for feedback
+
+With this in mind, I'd like to start now thinking through the steps on how to prepare. This design outlines some small steps. Additionally, this document is open to additional suggestions in what we can do.
+
+# Detailed design
+
+## Versioning Namespace Documents
+
+In addition to moving to 1.0, we need to actually start versioning namespace documents. The reason for this is that we do not want a breaking change in a namespace to require a major version bump for the entire spec.
+
+**Proposal**: Start versioning current and future namespaces. We can do this immediately.
+
+## Namespace Functionality
+
+The one area of Refract that we have not implemented in our implementations is namespaces. Currently, the design of namespaces mimics that of XML, but the way we have implemented it does not follow. In Refract, each element can have its own lexically-scoped elements as defined in its own namespace. This allows for namespaces to be defined and used deep within a data structure.
+
+Our implementations have diverted from this concept, and are rather relying on the document itself to define the elements includes. We still have namespaces, but the namespaces are not scoped within elements inside the document and are not explicitly defined in the documents.
+
+In order to get to 1.0, we need to think about whether we want to leave these kinds of things in the spec and change to them, or if we want to align the spec with what we've practically done.
+
+**Proposal**: Consider removing ability for namespaces to be lexically scoped on each element and make namespaces rather document-defined as we've done in our implementations.
+
+## Separating Out Serialization Formats
+
+An important concept of Refract is that it is a model for representing data structures and is not a JSON format. With that said, we do have two current formats as defined in the spec, and all examples in the spec use one of these formats.
+
+This decoupling I think is important, because it gives freedom to implementors to implement the best way they see fit in their language/platform, and it allows for other serialization formats to spring up that may be helpful to others.
+
+**Proposal**: Explain the difference in the conceptual model and the serialization formats, and potentially move these serialization descriptions to their own file in the spec repo, separating them from the main spec.
+
+# Drawbacks
+
+This could be considered to be premature. I am open to that discussion, and also to the idea of doing some of the things outlined above without having 1.0 in mind.
+
+# Alternatives
+
+We can extend the time we spend in pre-1.0 days.
+
+# Unresolved questions
+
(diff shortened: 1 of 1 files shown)


text/0000-refract-1.0.md
```diff
@@ -0,0 +1,56 @@
...
+# Summary
+
+The document provides some thoughts on how we can move forward to version 1.0 of Refract. This document is not proposing we move to 1.0 immediately, but to rather start putting steps in place to get us there quickly and being intentional about it.
+
+# Motivation
+
+As Refract is moving into production environments, we need to keep the following in mind:
+
+1. We want to have as a concise of a spec as necessary for newcomers
+1. We want to introduce breaking changes thoughtfully, which is what SemVer encourages. We currently can't do SemVer correctly without moving to 1.0.
```
kylef: 1.0.0 if we're using server :wink:.

text/0000-refract-1.0.md
```diff
@@ -0,0 +1,56 @@
...
+1. We want to start registering media types for Refract representations
+1. Moving this to production means people are actually going to start using this spec in the wild, and we need to prepare for feedback
+
+With this in mind, I'd like to start now thinking through the steps on how to prepare. This design outlines some small steps. Additionally, this document is open to additional suggestions in what we can do.
+
+# Detailed design
+
+## Versioning Namespace Documents
+
+In addition to moving to 1.0, we need to actually start versioning namespace documents. The reason for this is that we do not want a breaking change in a namespace to require a major version bump for the entire spec.
```
kylef: Since these are versioned separately, perhaps they belong in separate repositories so we can correctly tag releases.

smizell: @kylef You're probably right there. We can then link from the README accordingly. Sounds good to me :)

text/0000-refract-1.0.md
smizell: Ah, yes, I was lazy :) I'll change.

text/0000-refract-1.0.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Proposal**: Consider removing ability for namespaces to be lexically scoped on each element and make namespaces rather document-defined as we've done in our implementations.
+
+## Separating Out Serialization Formats
+
+An important concept of Refract is that it is a model for representing data structures and is not a JSON format. With that said, we do have two current formats as defined in the spec, and all examples in the spec use one of these formats.
+
+This decoupling I think is important, because it gives freedom to implementors to implement the best way they see fit in their language/platform, and it allows for other serialization formats to spring up that may be helpful to others.
+
+**Proposal**: Explain the difference in the conceptual model and the serialization formats, and potentially move these serialization descriptions to their own file in the spec repo, separating them from the main spec.
```
smizell: We could also have a separate repo for each of these, if necessary.

text/0000-refract-1.0.md
```diff
@@ -0,0 +1,56 @@
...
+
+# Drawbacks
+
+This could be considered to be premature. I am open to that discussion, and also to the idea of doing some of the things outlined above without having 1.0.0 in mind.
```
pksunkara, line 54: I agree with this being premature. We should let the spec evolve rather than trying to force 1.0.0

pksunkara:
I think we need to address the things mentioned in the RFC as part of 1.0.0. But I don't think we need to be hasty about it. I would like to have a bit more time and try refract in production first internally at Apiary so that we can change any breaking stuff first before moving to 1.0.0

For now, let's leave this PR as it is once everyone reviews it and keep adding to this if we come up with more stuff for 1.0.0

text/0000-refract-1.0.md
smizell, line 54: Just a note, I'm proposing we try to be intentional and determine steps we need to get to 1.0.0. I think we are actually a ways off from getting there though and not proposing we do it immediately. 

Even if a step is allow for this to be in production for a bit, I'm good with that. Just wanting us to think what the steps would be.

pksunkara, line 54: >  Just wanting us to think what the steps would be.

:+1:

text/0000-refract-1.0.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Namespace Functionality
+
+The one area of Refract that we have not implemented in our implementations is namespaces. Currently, the design of namespaces mimics that of XML, but the way we have implemented it does not follow. In Refract, each element can have its own lexically-scoped elements as defined in its own namespace. This allows for namespaces to be defined and used deep within a data structure.
+
+Our implementations have diverted from this concept, and are rather relying on the document itself to define the elements includes. We still have namespaces, but the namespaces are not scoped within elements inside the document and are not explicitly defined in the documents.
+
+In order to get to 1.0.0, we need to think about whether we want to leave these kinds of things in the spec and change to them, or if we want to align the spec with what we've practically done.
+
+**Proposal**: Consider removing ability for namespaces to be lexically scoped on each element and make namespaces rather document-defined as we've done in our implementations.
```
smizell: This is currently being proposed in https://github.com/refractproject/rfcs/pull/22 as a way to move toward using hyperlinking instead of namespacing like this.

This does not mean we remove the current namespaces. It does mean that the namespace functionality in the current spec will be removed.

smizell:
There are three things in this spec to accomplish. Namespacing is being discussed in another RFC and the other two are simply reorganizing some files. I propose we do these three things, yet still wait on moving completely to 1.0.0. 

If that works, I'll rework this RFC a bit to align with that idea and we can work on these few changes.

text/0000-refract-1.0.md
kylef: Did #30 address this for you @smizell?

kylef: Also saw #31.

smizell: Yes, I think #30 resolves this. Will update.

text/0000-refract-1.0.md
```diff
@@ -0,0 +1,56 @@
...
+
+With this in mind, I'd like to start now thinking through the steps on how to prepare. This design outlines some small steps. Additionally, this document is open to additional suggestions in what we can do.
+
+# Detailed design
+
+## Versioning Namespace Documents
+
+In addition to moving to 1.0.0, we need to actually start versioning namespace documents. The reason for this is that we do not want a breaking change in a namespace to require a major version bump for the entire spec.
+
+**Proposal**: Start versioning current and future namespaces and move each namespace to their own repo. We can do this immediately.
```
smizell: I have moved the discussion to here:
https://github.com/refractproject/rfcs/pull/34

pksunkara:
@smizell This needs to be updated.

pksunkara:
The serialisation has been done. And there is an issue tracking the inheritance on the Refract Spec. So, should we close this PR?

PR #14, opened 2015-08-29
Add Embedded Refract serialization

smizell:
Proposal for thinking about a new serialization format.

### DIFF
--- a/text/0000-embedded-refract.md
+++ b/text/0000-embedded-refract.md
@@ -1,0 +1,361 @@
+- Start Date: 2015-08-29
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This document outlines a new serialization format for embedding Refract.
+
+# Motivation
+
+1. Our current formats either look like XML, the DOM, or Lisp
+1. People already "get" JSON
+1. Other serializations we have are verbose
+
+# Detailed design
+
+## Overview
+
+For this serialization, there are two basic concepts:
+
+1. There is a reserved property name of `refract`
+1. There is a Refract Object with properties:
+    - element
+    - meta
+    - attributes
+    - content
+
+All valid JSON that adheres to reserved property is valid Embedded Refract.
+
+## Basic JSON
+
+We'll start with a basic JSON example, which we'll say is a person and their bowling scores.
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": 28,
+  "scores": [150, 202, 145]
+}
+```
+
+## Refracting Object Elements
+
+In Refract, this is an object with four members. With Embedded Refract, I can annotate by embedding refract using the `refract` property.
+
+```json
+{
+  "refract": {
+    "element": "object",
+    "meta": {
+      "id": "bowler-103"
+    }
+  },
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": 28,
+  "scores": [150, 202, 145]
+}
+```
+
+This is what this would like in full Refract.
+
+```json
+{
+  "element": "object",
+  "meta": {
+    "id": "bowler-103"
+  },
+  "content": [
+    {
+      "element": "member",
+      "content": {
+        "key": {
+          "element": "string",
+          "content": "first_name"
+        },
+        "value": {
+          "element": "string",
+          "content": "John"
+        }
+      }
+    },
+    {
+      "element": "member",
+      "content": {
+        "key": {
+          "element": "string",
+          "content": "first_name"
+        },
+        "value": {
+          "element": "string",
+          "content": "Doe"
+        }
+      }
+    },
+    {
+      "element": "member",
+      "content": {
+        "key": {
+          "element": "string",
+          "content": "age"
+        },
+        "value": {
+          "element": "number",
+          "content": 28
+        }
+      }
+    },
+    {
+      "element": "member",
+      "content": {
+        "key": {
+          "element": "string",
+          "content": "scores"
+        },
+        "value": {
+          "element": "array",
+          "content": [
+            {
+              "element": "number",
+              "content": 150
+            },
+            {
+              "element": "number",
+              "content": 202
+            },
+            {
+              "element": "number",
+              "content": 145
+            }
+          ]
+        }
+      }
+    }
+  ]
+}
+```
+
+## Refracting Value Elements
+
+If values do not have any meta values or attributes, there is no need to represent them as refracted. If I were to add some annotation to one of the values, it would show up as Refract and would be defined as Refracted by use of the `refract` property keyword.
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": {
+    "refract": {
+      "element": "number",
+      "meta": {
+        "id": "bowner-103-age"
+      },
+      "content": 28
+    }
+  },
+  "scores": [150, 202, 145]
+}
+```
+
+Above, I've added an ID to the age of the bowler. Notice how I just refracted that value in that specific place. This is also the same as:
+
+```json
+{
+  "refract": {
+    "element": "object",
+    "meta": {
+      "id": "bowler-103"
+    },
+    "content": {
+      "age": {
+        "element": "number",
+        "meta": {
+          "id": "bowner-103-age"
+        },
+        "content": 28
+      }
+    }
+  },
+  "first_name": "John",
+  "last_name": "Doe",
+  "scores": [150, 202, 145]
+}
+```
+
+This will also support full member elements, so even this below is equivalent (note that it's a slight modification from the full Refract):
+
+```json
+{
+  "refract": {
+    "element": "object",
+    "meta": {
+      "id": "bowler-103"
+    },
+    "content": [
+      {
+        "element": "member",
+        "content": {
+          "key": "age",
+          "value": {
+            "refract": {
+              "element": "number",
+              "meta": {
+                "id": "bowner-103-age"
+              },
+              "content": 28
+            }
+          }
+        }
+      }
+    ]
+  },
+  "first_name": "John",
+  "last_name": "Doe",
+  "scores": [150, 202, 145]
+}
+```
+
+In full Refract, you would have to also provide the element for the key "age" in the previous example, but since we can rely on the `refract` keyword, we can leave it unrefracted. We could Refract it more later if necessary.
+
+## Refracting Array and Array Items
+
+Array items can be embedded without refracting the entire array.
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": 28,
+  "scores": [
+    {
+      "refract": {
+        "element": "number",
+        "meta": {
+          "id": "bowler-103-game-1"
+        },
+        "content": 150
+      }
+    },
+    202,
+    145]
+}
+```
+
+Full arrays can also be refracted.
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": 28,
+  "scores": {
+    "refract": {
+      "element": "array",
+      "meta": {
+        "id": "bowler-103-scores"
+      },
+      "content": [150, 202, 145]
+    }
+  }
+}
+```
+
+## Refracting Attributes
+
+This makes it nice for refracting attributes, because now you know what attributes have been refracted. This could become:
+
+```json
+{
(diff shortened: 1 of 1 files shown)


smizell:
I also wanted to point out that I'm interested in this for the sake of adding hypermedia to existing JSON structures.

``` json
{
  "users": [
    {
      "id": 1,
      "name": "John Doe",
      "email": "john.doe@example.com"
    },
    {
      "id": 2,
      "name": "Jane Doe",
      "email": "jane.doe@example.com"
    }
  ]
}
```

becomes:

``` json
{
  "users": [
    {
      "refract": {
        "element": "resource",
        "attributes": {
          "rels": ["user"],
          "href": "/users/1"
        }
      },
      "id": 1,
      "name": "John Doe",
      "email": "john.doe@example.com"
    },
    {
      "refract": {
        "element": "resource",
        "attributes": {
          "rels": ["user"],
          "href": "/users/2"
        }
      },
      "id": 2,
      "name": "Jane Doe",
      "email": "jane.doe@example.com"
    }
  ]
}
```

Or without the transcluded data:

``` json
{
  "users": [
    {
      "refract": {
        "element": "resource",
        "attributes": {
          "rels": ["user"],
          "href": "/users/1"
        }
      }
    },
    {
      "refract": {
        "element": "resource",
        "attributes": {
          "rels": ["user"],
          "href": "/users/2"
        }
      }
    }
  ]
}
```

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+## Overview
+
+For this serialization, there are two basic concepts:
+
+1. There is a reserved property name of `refract`
+1. There is a Refract Object with properties:
+    - element
+    - meta
+    - attributes
+    - content
```
danielgtaylor, line 26: Is there any guidance on how to handle name collisions? What would be the expected behavior of an implementation that could not produce this serialization because of a pre-existing `refract` property?

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+## Basic JSON
+
+We'll start with a basic JSON example, which we'll say is a person and their bowling scores.
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": 28,
+  "scores": [150, 202, 145]
```
danielgtaylor, line 39: I think we need to list out the mappings from JSON types to element names for the auto-conversion that happens here.

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+```
+
+## Refracting Object Elements
+
+In Refract, this is an object with four members. With Embedded Refract, I can annotate by embedding refract using the `refract` property.
+
+```json
+{
+  "refract": {
+    "element": "object",
```
danielgtaylor, line 50: Is `element` required? Seems like it should be optional here.

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+    "meta": {
+      "id": "bowler-103"
+    },
+    "content": [
+      {
+        "element": "member",
+        "content": {
+          "key": "age",
```
danielgtaylor, line 199: I don't have a concrete example, but this is raising a red flag for me since it may lead to ambiguous cases where we cannot determine if a value is refracted. The first element of the first `content` array has no `refract` key and so an implementer must determine whether it is refracted by the `element` and `content` keys instead. I'm worried about this.

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": {
+    "refract": {
+      "element": "number",
+      "content": 28
+    },
+    "foo": "bar"
```
danielgtaylor, line 324: We may want to disallow this if it seems confusing enough and just treat it as an error. There's already a lot of magic going on :wink:

danielgtaylor:
This is definitely an interesting idea. I like the concept, but we should make sure it isn't too confusing.

In my opinion this also raises the question of which formats should be supported by a given implementation and whether the spec can provide some guidance on that issue.

text/0000-embedded-refract.md
smizell, line 26: Hmm, I'm not sure exactly. We could make it less likely to collide, like with `_refract`?

text/0000-embedded-refract.md
smizell, line 50: It could be optional for all object elements, but it is needed for all other element types otherwise.

text/0000-embedded-refract.md
smizell, line 199: So, inside the `refract` object, it's a totally different world :) For instance, the content of an object element can either be an object itself or an array of member elements (which is how it currently is). 

BUT we can change up how we manage the member's content to not have to refract everything there. For instance, the `key` value can be a JSON primitive or it could be a special refract object itself. Oh the recursion!!

text/0000-embedded-refract.md
smizell, line 324: I'm good with that. All properties would then be ignored :)

smizell:
@danielgtaylor My hope would be that you could add Refract to existing JSON, especially adding hypermedia to non-hypermedia structures. It would be very similar to JSON-LD, but a lot more powerful in that you could do lots of different types and stuff.

Regarding implementations, I would like to see this kind of stuff be a plugin for minim or some added on library. So we could keep the full version as the default, then include these other formats with `use`. It would go through all base types and add methods for `toWhatever` or however we wanted to do it.

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+  },
+  "scores": [150, 202, 145]
+}
+```
+
+I would suggest it be treated as an attribute. This would be the same as:
```
pksunkara: I don't think we should do this.

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+```json
+{
+  "refract": {
+    "element": "object",
+    "meta": {
+      "id": "bowler-103"
+    },
+    "attributes": {
+      "foo": {
+        "refract": {
```
pksunkara: Wait what? I know the value of `foo` can be a refract element. But we don't need the `refract` keyword and object here.

``` json
"foo": {
  "element": "string",
  "content": "baz"
}
```

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+```
+
+Full arrays can also be refracted.
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": 28,
+  "scores": {
```
pksunkara, line 252: I am not asking for any changes here but I am really worried about the parser or whatever which deals with this. `scores` is an array but if it is refracted, it is an object.

text/0000-embedded-refract.md
```diff
@@ -0,0 +1,361 @@
...
+
+```json
+{
+  "first_name": "John",
+  "last_name": "Doe",
+  "age": {
+    "refract": {
+      "element": "number",
+      "meta": {
+        "id": "bowner-103-age"
```
pksunkara: Did you mean `bowler`? Similarly below.

text/0000-embedded-refract.md
pksunkara, line 26: I wanted to propose `_refract`. I really don't like `refract`. It feels like we are taking away user's freedom. But using `_refract` diminishes it a bit.

smizell, line 26: I will change to `_refract` then. Works for me!

text/0000-embedded-refract.md
smizell, line 252: I actually had a branch that handled this, and it wasn't bad. The only thing you do is check every object for `_refract`. If it's there, you convert to element differently. So:
- If object, check for `_refract`
- If `_refract` is there, use special instructions for converting the Refract Object to element instance
- If not there, treat as normal JSON and convert to literal element instance

text/0000-embedded-refract.md
smizell: BUT, since we have this new rule in here we can benefit from it. This means I can know the signature of a refract element anywhere in the attribute tree, which means I don't need to rely on special instructions on what needs to be refracted and what doesn't. The representation tells me with `_refract`.

If you do the example you gave, that means we need something separate that says "Hey, `foo` is already refracted," but if we have `_refract`, it's right there in the data.

The only other option is to say, "Does this object have element and content? If so, it's refracted." We don't want a rule like that because that really hinders the user's freedom.

text/0000-embedded-refract.md
smizell: I agree now :)

https://github.com/refractproject/rfcs/pull/14/files#r38401396

text/0000-embedded-refract.md
pksunkara: The main issue I have here is that there is embedded refract in refract element. I don't think we should do that.

pksunkara:
Just one [issue](https://github.com/refractproject/rfcs/pull/14#discussion_r38700608) I am having concerns about. Embedded refract is a good thing. But embedded refract in refract element? How deep does the rabbit hole go?

text/0000-embedded-refract.md
smizell: Well, normal Refract can be in normal Refract, so this would be a problem for normal Refract. The issue is there already, but this actually resolves part of the issue of not knowing what is/isn't refracted.

Is your concern the recursion?

smizell:
Comments addressed:
- Fixed typos
- Changed to `_refract`

pksunkara:
@smizell Continuing the discussion here.

> Is your concern the recursion?

It's not about the recursion. It's not about the implementation or anything. It's about concepts. It's more about the fact that refract is a core thing and embedded refract is concept built on it. What use is there for embedded refract to be in refract? Refract already allows elements in refract. If we try to use embedded refract in refract, it does nothing different except maybe add an extra nested level of object.

smizell:
It actually does add something in that you know what has or has not been refracted. It tells you more that normal refract. No need for element specific code such as we have in Minim. 

See https://github.com/refractproject/minim/blob/master/lib/primitives/base-element.js and `convertAttributesToElements`. 

I'm ok with not including it, but I do think you actually could gain a lot.

pksunkara:
From your example

``` json
{
  "_refract": {
    "element": "object",
    "meta": {
      "id": "bowler-103"
    },
    "attributes": {
      "foo": {
        "_refract": {
          "element": "string",
          "content": "baz"
        }
      }
    }
  },
  "first_name": "John"
}
```

If we don't use embedded refract in refract, this is how the example looks.

``` json
{
  "_refract": {
    "element": "object",
    "meta": {
      "id": "bowler-103"
    },
    "attributes": {
      "foo": {
        "element": "string",
        "content": "baz"
      }
    }
  },
  "first_name": "John"
}
```

I don't see any benefit in the first example when compared to the second one. Actually the extra `_refract` makes it more complex.

> I do think you actually could gain a lot.

Can you give me a small example? Maybe I am having problems visualising it.

smizell:
Well, let's use your example first. 

``` json
{
  "_refract": {
    "element": "object",
    "meta": {
      "id": "bowler-103"
    },
    "attributes": {
      "foo": {
        "element": "string",
        "content": "baz"
      }
    }
  },
  "first_name": "John"
}
```

How would a client know that `foo` has been refracted? You would need some domain-specific, out-of-band knowledge to know it. The only other option is to say that "You MAY determine an object is refracted by looking for the presence of an `element`." But we don't want that.

So, we can get around it by treating the value of attributes as a normal object, and allow the already-reserved `_refract` property to tell the parser it has already been refracted.

``` json
{
  "_refract": {
    "element": "object",
    "meta": {
      "id": "bowler-103"
    },
    "attributes": {
      "foo": {
        "_refract": {
          "element": "string",
          "content": "baz"
        }
      }
    }
  },
  "first_name": "John"
}
```

In this example, the parser would see `_refract` and know it's already refracted without any other knowledge. 

---

The benefit here is that you can have refracted elements mixed into the attributes tree anywhere, just like with a normal object. It would require no out-of-band knowledge or special parsing instructions like [this](https://github.com/refractproject/minim-api-description/pull/1/files#diff-1071bd6aebf6be905eb91db38be81918R7).

pksunkara:
Understood now. Thanks :smile:

PR #17, opened 2015-10-16
RFC for serialization

danielgtaylor:
The goal of this RFC is to formalize and document the various serialization formats used by Refract, as well as explicitly list how to handle ambiguity in refracted metadata and attributes.

### DIFF
--- a/text/0000-serialization.md
+++ b/text/0000-serialization.md
@@ -1,0 +1,187 @@
+- Start Date: 2015-10-16
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+The goal of this RFC is to formalize and document the various serialization formats used by Refract, as well as explicitly list how to handle ambiguity in refracted metadata and attributes.
+
+# Motivation
+
+1. Currently they are not defined in the specification. There is one RFC for Embedded Refract but it is not visible to people looking at the Refract specification repository.
+2. Optionally refracted element metadata and attributes cause ambiguity issues with some serialization formats. We need to formally document how to avoid these ambiguities.
+3. Doing this could help us to remove some domain and element-specific knowledge from the parser because it removes ambiguity. For example, in the API Description namespace a resource may have HTTP headers as an attribute, which is always an element. If we remove the ambiguity we can remove special logic for this attribute and treat it like any other, in this case seeing it is refracted and loading the `httpHeaders` element.
+
+# Detailed design
+
+At the time of this writing there are three serialization formats in use to save and load Refract structured data. They are:
+
+- Full JSON
+- Compact (list-based) JSON
+- Embedded
+
+The [Embedded Refract](https://github.com/refractproject/rfcs/blob/master/text/0005-embedded-refract.md) RFC contains information about when and how to make element metadata and attributes into refract elements, and explains how the parser knows how to handle them. This RFC will focus on the other two formats, which should be described in the Refract specification.
+
+## Full JSON
+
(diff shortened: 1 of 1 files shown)


text/0000-serialization.md
```diff
@@ -0,0 +1,187 @@
...
+          "element": "string",
+          "content": "foo"
+        }
+      }
+    }
+  ]
+}
+
+// Unambiguous compact JSON version
+['array', {}, {}, ['foo', {}, {}, 0]]
```
smizell, line 171: A couple of thoughts on these examples.

The full version is simply refracting the element as a normal object. I believe we would do this anyway within Minim, and would treat it as a normal Object Element. 

Regarding the compact example, I don't believe this is a valid array element. I believe it would have to be:

``` js
['array', {}, {}, [
  ['string', {}, {}, 'foo'],
  ['object', {}, {}, []],
  ['object', {}, {}, []],
  ['number', {}, {}, 0]
]]
```

We would still do this in any situations where Compact Refract was not expected I believe.

So I actually believe, without any domain-specific knowledge, we will actually already do as you've defined. The difficulty is knowing where that domain-specific knowledge is in the tree.

text/0000-serialization.md
```diff
@@ -0,0 +1,187 @@
...
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+The goal of this RFC is to formalize and document the various serialization formats used by Refract, as well as explicitly list how to handle ambiguity in refracted metadata and attributes.
+
+# Motivation
+
+1. Currently they are not defined in the specification. There is one RFC for Embedded Refract but it is not visible to people looking at the Refract specification repository.
```
smizell, line 11: I believe we should consider the Full version of Refract the main Refract serialization format, and then move the other versions to their own repositories as their own specs. This would mean we'd move the Compact Refract to its own repo along with the Embedded Refract.

That would mean we'd take your recommendations here for Compact, use the RFC for Embedded, and leave the Full version alone. Thoughts?

text/0000-serialization.md
smizell, line 171: I would like to also explore how we could tell the parse that these are actual elements. We've talked about a few options here, but just wanted to make this comment. :)

smizell:
@danielgtaylor I've added some thoughts.

I'd also like to think about [this comment](https://github.com/refractproject/rfcs/pull/17#discussion_r45826181) where we use your work on the compact version and the RFC for the embedded and create separate repositories for each. We would then remove the compact version from the base Refract spec.

text/0000-serialization.md
```diff
@@ -0,0 +1,187 @@
...
+2. Optionally refracted element metadata and attributes cause ambiguity issues with some serialization formats. We need to formally document how to avoid these ambiguities.
```
kylef, line 12: Which serialization formats are these ambiguous in?

smizell, line 12: @danielgtaylor Will this be resolved with this?

https://github.com/refractproject/rfcs/pull/29

If so, would you care to update this RFC accordingly?

smizell:
How do you all feel about pulling the compact version out of the base spec? I'm not proposing we get rid of it, but I'm wondering if people need to implement it in order to fulfill the base Refract spec. Thoughts?

Compact Refract is only mentioned in passing in the spec, and is not explained in detail (like this RFC aims to do).

zdne:
> How do you all feel about pulling the compact version out of the base spec? I'm not proposing we get rid of it, but I'm wondering if people need to implement it in order to fulfill the base Refract spec. Thoughts?

Let's remove it from there

smizell:
See https://github.com/refractproject/refract-spec/pull/53 for moving Compact Refract

PR #18, opened 2015-11-18
RFC for media types for each namespace

smizell:
This proposes media types so all users can use the same media types when communicating.


netmilk:
:+1:

danielgtaylor:
LGTM :ship:-it!

pksunkara:
:+1:

PR #20, opened 2015-11-20
RFC for parse result meta data

smizell:
This may be a little too aggressive for this RFC, but I have provided alternatives.

I personally like the idea of being able to add links to any element to allow the user to explore additional meta data. This allows anything to become a hyperlink in Refract, which could be very powerful in other situations.

### DIFF
--- a/text/0000-parse-result-meta-data.md
+++ b/text/0000-parse-result-meta-data.md
@@ -1,0 +1,78 @@
+- Start Date: 2015-11-20
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This RFC outlines ideas in adding meta data into into elements through the
+means of meta links and adding link relations to the Parse Result namespace.
+
+# Motivation
+
+Parsing happens in multiple steps and information is added by different layers
+in the stack in some situations. It would be helpful for end users to have meta
+data to let the user know where data came from and which information was added
+that did not exist in the original representation.
+
+The idea with proposing meta links would be to provide a generic way to link to
+out-of-band information to allow the user to explore more information in a
+given context.
+
+# Detailed design
+
+The proposal is to add a `links` property to `meta` and propose two link
+relations for the Parse Result namespace.
+
+## Links Property in Meta
+
+The property would be defined as the following the `meta` object of all
+elements. This would be added in the Refract base spec.
+
+- `links` (array)
+    - (object)
+        - `rel` - Link relation of the link
+        - `href` - Resolvable URL for the link
+
+The `rel` property MAY be any defined [link
+relation](http://www.iana.org/assignments/link-relations/link-relations.xhtml)
+or a resolvable URL to documentation on the link relation.
+
+## Parse Result Link Relations
+
+### `inferred`
+
+The `inferred` link relation defines a link to documentation on how an element
+was inferred. The presence of an `inferred` link tells the user that the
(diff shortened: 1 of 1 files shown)


text/0000-parse-result-meta-data.md
```diff
@@ -0,0 +1,78 @@
...
+
+# Motivation
+
+Parsing happens in multiple steps and information is added by different layers
+in the stack in some situations. It would be helpful for end users to have meta
+data to let the user know where data came from and which information was added
+that did not exist in the original representation.
+
+The idea with proposing meta links would be to provide a generic way to link to
+out-of-band information to allow the user to explore more information in a
```
fosrias: Not out of band if you can link to it. Maybe "link to what would normally be out-of-band ..."

text/0000-parse-result-meta-data.md
```diff
@@ -0,0 +1,78 @@
...
+
+The `rel` property MAY be any defined [link
+relation](http://www.iana.org/assignments/link-relations/link-relations.xhtml)
+or a resolvable URL to documentation on the link relation.
+
+## Parse Result Link Relations
+
+### `inferred`
+
+The `inferred` link relation defines a link to documentation on how an element
```
fosrias, line 48: What would be the purpose of such documentation? I.e. why would it exist?

text/0000-parse-result-meta-data.md
smizell: Good point :) Will fix.

text/0000-parse-result-meta-data.md
smizell, line 48: To go to the world of hypermedia representors, say you were to parse Collection+JSON to a canonical model. There are some situations where you would either drop information or make assumptions/inferences. This would be a link from the elements that have been inferred to a document describing why this is so and what was inferred.

Of course, a client accessing this document can tell if an element has been inferred by the presence of this link relation, but it can also access additional information if need be through the link.

text/0000-parse-result-meta-data.md
```diff
@@ -0,0 +1,82 @@
...
+elements. This would be added in the Refract base spec.
+
+- `links` (array)
+    - (object)
+        - `rel` - Link relation of the link
+        - `href` - Resolvable URL for the link
+
+The `rel` property MAY be any defined [link
+relation](http://www.iana.org/assignments/link-relations/link-relations.xhtml)
+or a resolvable URL to documentation on the link relation.
```
zdne, line 38: "or" vs. "and/or" (is `rel` allowed with `href`) ?

zdne:
@smizell  I like it in general albeit I feel this RFC is trying to do two things: 
1. Define the framework for providing additional metadata on element
2. Defining certain link relations

Not saying this is bad, per se. But maybe we want to allow ANY (not necessary parsing-related) metadata in the form of `Links` object and THEN define link relations specific for parsing. Just my few cents.

danielgtaylor:
@smizell I'm having some trouble understanding what exactly this is accomplishing for us as it's all very abstract and feels like hypermedia-magic. Is it possible to add a real-world example either here in the comments or to the RFC of how we expect it to be used and why it makes sense to add this `links` property?

I understand the `inferred` attribute or class, just having some trouble visualizing the `links` concept. At least to me, saying that something is inferred isn't really a link relationship.

smizell:
Closing. Will open a new PR that is only for hyperlinking.

PR #22, opened 2015-11-24
RFC for Adding Hyperlinking

smizell:
This adds the RFC for adding hyperlinking into the base Refract specification.

### DIFF
--- a/text/0000-add-hyperlinks.md
+++ b/text/0000-add-hyperlinks.md
@@ -12,14 +12,12 @@ As we encounter more real-world Refract situations, we are noticing that we seem
 
 # Design
 
-This proposal is for adding a `links` property to the `meta` object in Refract elements.
+This proposal is for adding a Link Element to the base spec.
 
-- `links` (array)
-    - (object)
-        - `rel` - Link relation of the link
-        - `href` - Resolvable URL for the link
-
-The `rel` property SHOULD conform to [RFC 5988](https://tools.ietf.org/html/rfc5988). Link relations should be as registered or extension link relations, and MAY be defined in profile documents.
+- `element`: link (string, fixed)
+- `attributes`
+    - `relation` (string) - Link relation type as specified in [RFC 5988][].
+    - `href` (string) - The URI for the given link
 
 ## Use Cases
 
@@ -39,12 +37,18 @@ This example shows a `foo` element with two links. The presence of the `edit` li
   meta: {
     links: [
       {
-        rel: 'edit',
-        href: '/bar/1'
+        element: 'link',
+        attributes: {
+          relation: 'edit',
+          href: '/bar/1'
+        }
       },
       {
-        rel: 'self',
-        href: '/bar/1'
+        element: 'link',
+        attributes: {
+          relation: 'self',
+          href: '/bar/1'
+        }
       }
     ]
   }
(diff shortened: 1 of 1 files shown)


danielgtaylor:
LGTM, but would love to get feedback from @almad and @zdne

text/0000-add-hyperlinks.md
```diff
@@ -0,0 +1,73 @@
...
+# Design
+
+This proposal is for adding a `links` property to the `meta` object in Refract elements.
+
+- `links` (array)
+    - (object)
+        - `rel` - Link relation of the link
+        - `href` - Resolvable URL for the link
+
+The `rel` property MAY be any defined [link relation](http://www.iana.org/assignments/link-relations/link-relations.xhtml), a resolvable URL to documentation on the link relation, or defined in a `profile` link.
```
kylef: I'm not sure, what does "resolvable" in this context mean?

zdne:
@smizell this is draft is HUGE in its scope. 

Please let me break it down for having a better grasp on the problem. Of what this draft has to offer:
1. Express what relation element has through links
2. Embed (transclude) links into the element's meta
3. Remove refract namespaces and replace them with `profile` relations

For the simplicity and speed I would like to ask you for following:
1. Do not solve embedding of links at all. (No.2)
2. Do not solve replacing namespaces with `profile` (No.3) at this point (albeit I would love to do this).

Furthermore I would like to see a synergy with the `transition` element defined in the API Description namespace ([Transition (Element)](https://github.com/refractproject/refract-spec/blob/master/namespaces/api-description-namespace.md#transition-element)) especially considering its `relation` property. At least I am looking for aligning the names (`rel` vs. `relation`). But I would appreciate consideration of reusing the link element in the relation element. 

As for removing namespaces – I am very much up for it, I would leave up to you and @netmilk to consider if this is something you want to get in to as part of current operations. 

Note Data Structure namespace **depends** on current base spec `link` element's `path` attribute and this is something that should be considered. 

But again I would prefer to have this (removing namespaces and using `profile` relation) as a separate discussion – otherwise we are trying to do too many things at once.

smizell:
@zdne Sorry for the confusion. I did not want to propose those changes in this RFC, but simply make them known so we can think ahead while thinking short term. Would hate to make the wrong decision for being shortsighted. Please simply keep those two additional ideas (embedding and removing namespacing) in the back of your mind as we proceed.

Regarding synthesizing with the Transition Element, could you explain this a little more? Are you meaning for this proposal to use a Transition Element instead of an array of objects with `rel` and `href`? If not, could you elaborate?

---

This is out of scope of this RFC, but I did want to add a note.

> Note Data Structure namespace depends on current base spec link element's path attribute and this is something that should be considered.

I have started a [preliminary branch](https://github.com/refractproject/refract-spec/blob/smizell/address-rfcs/refract-spec.md) for all of these changes. We would not remove `path` from the `ref` element for this reason.

zdne:
> Regarding synthesizing with the Transition Element, could you explain this a little more? Are you meaning for this proposal to use a Transition Element instead of an array of objects with rel and href? If not, could you elaborate?

Ideally I would love the same thing be called the same. (rel vs relation)

text/0000-add-hyperlinks.md
```diff
@@ -0,0 +1,84 @@
...
+
+In some situations, it may be advantageous to embed the resources themselves that would be found when the link was resolved. This current proposal is not addressing this yet, though it may be something we should look into in this RFC. This would happen if the design would go better together.
+
+## Additional Benefits
+
+Additionally, if hyperlinks are used in this way, we may be able to remove the concept of namespaces from the base Refract specification. The namespace properties allowed for programmatically including additional elements, but so far no library is using or has implemented this functionality. Since we've been using content types for extending Refract documents, we could use the `profile` link relation to define additional functionality and link relations.
+
+If we go this route, we can make the following changes to the base Refract specification.
+
+- Remove `prefix` and `namespace` from the `meta` object
```
fosrias: The point of this is? You plan to use links for namespaces or what?

smizell:
I have changed the link per the idea of @zdne. Basically, instead of a normal object, I am using a special Link Element that can be used as the basis for the Transition element.

Also, this opens the way for embedding responses through existing things in the API Description format. We'll leave that discussion until later, but in using an element, we provide a way for this to be extended very easily to either use `Asset` or `Response`

text/0000-add-hyperlinks.md
smizell: I am removing this from the current RFC discussion, but some thoughts (feel free to continue questions/thoughts here).

The current usage and implementation of Refract completely ignore everything dealing with namespacing. In adding hyperlinking, we can link to a `profile` that defines the usage for a given structure, which is what we're currently doing.

To actually make namespacing work, we will need to defined way to not only define namespaces (which we don't have nor will we have for a long while) but also a way to compare a given namespace with the actual data. Lots of work here to do before this could ever be supported.

Because there are no implementations or specifications even using this, and because this would require a great deal of work, and because this is a pre-1.0.0 version of things, we need to remove any cruft we can, and therefore we are thinking to remove namespacing.

What are your thoughts?

smizell:
Note that I also opened a PR for these changes in the spec repository. If things look OK with this, we can move to discussing changing the spec, which hopefully that PR will do.

zdne:
:+1: merging

PR #23, opened 2015-11-24
RFC for parse result link relations

smizell:
This is an RFC for link relations for the parse result namespace.

### DIFF
--- a/text/0000-parse-result-link-relations.md
+++ b/text/0000-parse-result-link-relations.md
@@ -1,0 +1,39 @@
+- Start Date: 2015-11-24
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


danielgtaylor:
Distilling my comments from our Slack conversation:
1. I am still a little uncomfortable with the `inferred` link relation. It just doesn't feel like a link to me, but I won't block this because I don't have a better solution that is as generic as this one. A property on meta called `inferred` is very specific to this particular use case.
2. The `origin` relation makes sense to me since it is a link to the originating software. I like this.

So, given we get some other eyes on this (@almad, @zdne, etc.) then I am okay with merging.

zdne:
I am OK with this once https://github.com/refractproject/rfcs/pull/22 is resolved. 

@danielgtaylor 

>  It just doesn't feel like a link to me, 

I think it is good if the `inferred` points to the author (tool) of the inferred element, so I think it could (read should) be a link

zdne:
Merging with #22 being active

PR #26, opened 2015-11-27
`hrefVariables` within `httpRequest`

kylef:
This RFC proposes we add a `hrefVariables` attribute to the `httpRequest` and change `href` to become a Templated Href instead of a Href.

### DIFF
--- a/text/0000-httprequest-href-variables.md
+++ b/text/0000-httprequest-href-variables.md
@@ -1,0 +1,44 @@
+- Start Date: 2015-11-27
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes we add a `hrefVariables` attribute to the `httpRequest` and
+change `href` to become a `Templated Href` instead of a `Href`.
+
+# Motivation
+
+From the perspective of tools such as testing, mocking, proxy, routing and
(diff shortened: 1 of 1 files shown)


kylef:
If anyone is interested in where this came from and related discussion it can be found on an [RFC to API Blueprint](https://github.com/apiaryio/api-blueprint-rfcs/pull/9#discussion_r45968857).

text/0000-httprequest-href-variables.md
```diff
@@ -0,0 +1,44 @@
...
+> - `href` (Href) - A concrete URI for the request. The href SHOULD be
+>   inherited from a parent transition by expanding the href and
+>   hrefVariables if unset.
+
+These proposed changes would change the meaning of the `href` attribute
+to become:
+
+> - `href` (Templated Href) - URI Template for this HTTP request. The `href`
+>   and `hrefVariables` SHOULD be inherited from a parent transition.
+>   When `href` is set, you MAY also set `hrefVariables`.
```
zdne, line 29: ~~I believe you should be able to set `hrefVariables` even without the `href` setotherwise we will end up with always duplicating the `href`~~

zdne, line 29: Actually let's keep it simple – if a parameters are discussed in the request the `href` has to be set.

PR #27, opened 2015-12-03
Add RFC for adding sends and accepts attributes

smizell:


danielgtaylor:
I think this looks good, however I have a general comment about this and other features we may want to support. I think it would be nice to look at popular examples of use in the wild before we add new features to Refract. For this specific example, here is what I have found:
1. API Blueprint: no support
2. [Swagger](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#fixed-fields): provides `produces` and `consumes` as hints of what values _may_ be used in their respective headers and response payloads.
3. [RAML](http://docs.raml.org/specs/1.0/#raml-10-spec-the-root-of-the-document): no support, but provides a default `mediaType` for response payloads that can be defined in the root of the document. Not the same feature, but probably related. Also worth noting that a feature like this can just be handled by a parser when creating refract assets with specific content types.

Also, if someone could comment on whether API Blueprint is considering this type of feature in the future it might be helpful to understand any ideas around that so we don't limit ourselves to a specific implementation that exists today. @zdne @kylef 

Anyway, I'm :+1: on this, so as long as other's sign off I say merge it in.

smizell:
Good point. I would vote for how Swagger currently does it where `sends` and `accepts` are merely hints.

Also, as a note, I went with `sends` and `accepts` because they focus more on the transaction, representation, and delivery rather than functionality, which is what `produces` and `consumes` do. May be subtle, but just wanted to mention the reason for it.

smizell:
Additionally, there SHOULD be examples for each hint.

text/0000-sends-accepts.md
```diff
@@ -0,0 +1,46 @@
+- Start Date: 2015-12-3
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This proposes adding `sends` and `accepts` attributes to the `Category` element
```
zdne, line 7: I am surprised to see the `category` here can you please elaborate?

text/0000-sends-accepts.md
```diff
@@ -0,0 +1,46 @@
...
+and `Transaction` element in the API Description format.
+
+# Motivation
+
+This accommodates the ability to capture what mime types an API or Transaction
```
zdne, line 12: Thinking more about this – I'd argue this should be nested under some sort of "mime-type" attribute.

text/0000-sends-accepts.md
```diff
@@ -0,0 +1,46 @@
...
+
+When either of these are not present, these values are determined by those in
+the HTTP Request and HTTP Response element. These values do not dictate what
+mime types a Transaction MUST use, but rather acts as hints.
+
+These attributes should be added to `Category` and `Transaction` elements.
+
+# Drawbacks
+
+This is adding information that is only for hinting, and may not actually
```
zdne, line 31: > we would need to encourage users to always provide examples when these attributes are used. 

that would be hard without providing a facility to this – for example an asset element in the accepts array (of tuples?)

smizell, line 31: I should say, we should encourage users to include a transaction for each given media type.

text/0000-sends-accepts.md
smizell, line 12: Such as:

``` js
{
  element: 'foo',
  attributes: {
    mimeTypes: {
      sends: [],
      accepts: []
    }
  }
}
```

text/0000-sends-accepts.md
smizell, line 7: Swagger supports this data at the [root level](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#swagger-object) as well as the [operation level](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operation-object).

zdne, line 7: @smizell I understand that. What I am grappling with is – are we just going to add "everything everywhere" just because format XYZ has it or are we after a **conceptually consistent data model**?

zdne, line 7: I am not saying this particular change is good or bad, rather I am missing the reasoning why we think it is conceptually good to have mime types attached to the category element.

text/0000-sends-accepts.md
zdne, line 12: perhaps unless you thinks it adds more burden

text/0000-sends-accepts.md
zdne, line 31: that (at this point) implies HTTP. We can (not saying we should) engineer this in a way that a mime type might come with an example (asset)

zdne:
For the record, at this point, I am :-1:  on this. Unless we identify the accepts / sends as conceptually correct elements (attributes) for the `category` and `transition` element.

As discussed with @smizell offline, we have to be really careful introducing new elements / attributes to this namespace _just to accommodate_ one format's data. We should be aspiring for higher goals in our quest to identify canonical API (description) elements.

pksunkara:
I feel we are being hasty on this. :-1: from me.

smizell:
I'm closing for now. I'm in agreement with @zdne and @pksunkara here.

PR #29, opened 2015-12-03
RFC to reserve keyword element.

smizell:


danielgtaylor:
Love it :+1:. This basically fixes #17 by formally making `element` reserved for serialized full refract.

> This means that no user can ever use the keyword `element`.

Actually, they can as long as the serializer is smart enough to refract that object into something like:

``` json
{
  "element": "object",
  "content": [
    {
      "element": "member",
      "content": {
        "key": "element",
        "value": "..."
      }
    }
  ]
}
```

It does make writing by hand and using that particular key more difficult though :smile:

smizell:
> Actually, they can as long as the serializer is smart enough to refract that object

You know, that is true. For instances of `Object Element`, if there is an "element" member, the entire object should be refracted. As for parsing, they would just do the same as you've mentioned. That seems easy to get around when serializing.

text/0000-reserve-element.md
```diff
@@ -0,0 +1,55 @@
...
+This RFC proposes to reserve `element` in the entire tree. This means that
+anywhere the `element` keyword is found, a parser SHOULD treat the object as a
+Refract Object.
+
+## Secondary Proposal
+
+This allows for us to know when an object is a Refract Object or not. It also
+allows for us only refract things that need to be refracted when serializing.
+In this situation, for example, the `key` property of a `Member Element` would
+not need to refracted if it is a string.
```
danielgtaylor, line 30: > ...things that need to be refracted when serializing.

We may want to define here exactly what that means. For example, we _must_ refract `member` elements or expect implementations to handle them some other way (in `object`'s content or in attributes like `hrefVariables` for example because it reintroduces domain-specific knowledge). Maybe a whitelist of types supported to not be refracted? Does that make sense?

smizell, line 30: So in thinking of this, I don't think we should pursue this too far. But it opens to lots of possibilities. Take this for example:

``` js
{
  foo: 'bar',
  baz: 'bab'
}
```

This _could_ be considered valid Refract if we reserve `element`. Now what if the "bar" value is some non-primitive element?

``` js
{
  foo: {
    element: 'non-primitive',
    content: 'bar'
  },
  baz: 'bab'
}
```

The only reason we would need to refract the entire object and use member elements here is if `foo` or `baz` have non-primitive elements or have meta or attributes.

---

My concern here is interop with current tooling. But in allowing us to optionally Refract things, it does make things a lot cleaner :)

---

> we must refract member elements

Given the example above, I'm not sure we have to, though interop with cooling tooling could require it. We would only need member elements if (a) the member element has meta or attributes or (b) a key in the object is non-primitive or has meta or attributes. Do you think this would still require a "MUST refract" here?

zdne:
Let's do it

text/0000-reserve-element.md
```diff
@@ -0,0 +1,55 @@
...
+Refract tree. Right now, we have to rely on something out of band to tell us
+when something MAY be refracted, but if we rely on the reserved keyword
+`element`, we can know if an object is a normal object or a Refract Object.
+
+Note: we are already doing this in Minim.
+
+# Design
+
+This RFC proposes to reserve `element` in the entire tree. This means that
+anywhere the `element` keyword is found, a parser SHOULD treat the object as a
```
pksunkara: anywhere the `element` key is present, right? Calling it keyword sounds weird and allows ambiguity of it being present in a value of an object.

text/0000-reserve-element.md
```diff
@@ -0,0 +1,55 @@
...
+Refract Object.
+
+## Secondary Proposal
+
+This allows for us to know when an object is a Refract Object or not. It also
+allows for us only refract things that need to be refracted when serializing.
+In this situation, for example, the `key` property of a `Member Element` would
+not need to refracted if it is a string.
+
+We can then follow this proposed rule when serializing: if an element is a
```
pksunkara, line 32: This rule would go into the RFC where we define serialisation rules.

pksunkara:
Minor comments, otherwise good to go.

text/0000-reserve-element.md
smizell, line 32: Added a note about this.

smizell:
@pksunkara comments addressed

text/0000-reserve-element.md
```diff
@@ -0,0 +1,56 @@
...
+primitive element and has no attributes or meta data, the element MAY be left
+as a primitive value. This rule SHOULD go into the RFC where we define
+[serialization rules](https://github.com/refractproject/rfcs/pull/17).
+
+# Drawbacks
+
+This means that no user can ever use the keyword `element`. We could provide
```
pksunkara: s/property/keyword

PR #30, opened 2015-12-03
Remove namespacing in base specification

smizell:


danielgtaylor:
I'm :+1: for removing namespaces in elements since they are not currently used, unless someone else has an reason they should stay (@zdne?). I think we should hold off on a formal way to do this unless/until we find a need for it.

text/0000-remove-namespacing.md
```diff
@@ -0,0 +1,50 @@
+- Start Date: 2015-12-03
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes we remove all namespacing in the Refract base specification
+while proposing we use `profile` links in its place.
+
+Note: this is not to be confused with the actual namespace documents we have
```
zdne, line 10: are they (namespace documents) going to be called profiles from now on?

text/0000-remove-namespacing.md
```diff
@@ -0,0 +1,50 @@
...
+- Remove `Namespacing` section in specification
+
+In its place, we will provide instructions for using the `profile` link
+relation for defining where the semantic definition for elements MAY be found.
+This `profile` link SHOULD conform to [RFC
+6906](https://www.ietf.org/rfc/rfc6906.txt).
+
+# Drawbacks
+
+Namespacing provides a very detailed way to extend documents and elements.
```
zdne, line 39: I would like to see some discussion about future compatibility. Eg. if we will need introduce namespaces alter in the future wont that be a breaking change?

text/0000-remove-namespacing.md
smizell, line 10: I think if we move to profiles, it seems like they should be called such, though I'm definitely open to suggestions here.

text/0000-remove-namespacing.md
smizell, line 39: I think the only thing that really isn't compatible is the current use of prefixes. Namespaces simply point to a document that defines semantics and other uses. Profiles can still exist on an element level. Prefixes are what allow for resolving collisions.

text/0000-remove-namespacing.md
zdne, line 10: I am +1 for using "profile"

PR #35, opened 2017-03-24
Fix type of meta and attributes

pksunkara:
Would love your input here @smizell


pksunkara:
For the record, I am actually okay with making the `meta` and `attributes` as an object even though it means keys can't have their own `meta` and `attributes`.

PR #37, opened 2017-03-27
Meta/Attributes Type

kylef:
This RFC offers an alternative to https://github.com/refractproject/rfcs/pull/35. After thinking about this and discussing with @smizell it seems there is no need to support having meta or attributes on an meta or attribute member/key itself and it would add additional complexity so it may not be worth supporting to keep Refract as simple as possible.


pksunkara (approved):

text/XXXX-meta-attributes-type.md
```diff
@@ -0,0 +1,50 @@
...
+the value is a Refract element. The value of meta or attributes MUST no longer
+be an array of member elements.
+
+## Drawbacks
+
+It will not be possible for elements declared in name spaces to allow
+meta/attributes to be stored on a member or member key. This would not effect
+any existing name-spaces however it may have impact in the future.
+
+Although this may impose a limitation, it is clearer to consumers of Refract.
```
smizell: I think this is a good point, and something to keep in mind from users as they give feedback. However, if a user _really_ needs to define this information, the prescribed way for now should be to wrap this in some object in attributes. For instance, here we're adding attributes to a member element in attributes, just not the top level.

```
{
    "element": "foo",
    "attributes": {
        "data": {
            "element": "object",
            "content": [
                {
                    "element": "member",
                    "attributes": {...}
                 }
             ]
        }
    }
}
```

smizell (approved):
Looks good to me!

text/XXXX-meta-attributes-type.md
pksunkara: They don't even need to wrap it. They can just add the meta/attributes to the value most of the times.

smizell: That's true. This would only be helpful in the very rare occasion (which may not exist) where a user wants to add meta/attributes on a member or member key on a root attribute. Surely that's a very, very special case that will never pop up.

PR #38, opened 2017-03-27
Element Pointer

kylef:
Element Pointer is currently a special case and the only type that does not extend from an element which means it has special rules when it comes to serialisation and de-serialisations. These have been brought up in the discussion at https://github.com/refractproject/rfcs/pull/36#discussion_r108148908 and while we are at it, I think it makes sense for the element pointer to become an element.

### DIFF
--- a/text/XXXX-element-pointer.md
+++ b/text/XXXX-element-pointer.md
@@ -1,0 +1,33 @@
+- Start Date: 2017-03-28
+- RFC PR: (leave this empty)
+- Refract Issue: (leave this empty)
+
+# Element Pointer
+
+## Summary
+
+Element Pointer is currently a special case and the only type that does not
+extend from an element which means it has special rules when it comes to
+serialisation and de-serialisations.
+
+## Motivation
+
+To aid simplicity and to remove complexity of Refract serialisation, we should
+move Element Pointer to be an element itself. This would be more consistent
+with other elements such as the Link element.
+
+## Detailed design
+
+The Element Pointer should become an Element Pointer element which can be
+described as follows:
+
+### Element Pointer
+
++ element: elementPointer (fixed)
++ attributes
+    + path (enum) - Path of the referenced element to transclude instad of element itself
+        + element (default)
+        + meta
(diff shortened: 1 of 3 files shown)


text/XXXX-element-pointer.md
```diff
@@ -0,0 +1,33 @@
...
+
+The Element Pointer should become an Element Pointer element which can be
+described as follows:
+
+### Element Pointer
+
++ element: elementPointer (fixed)
++ attributes
+    + path (enum) - Path of the referenced element to transclude instad of element itself
+        + element (default)
```
pksunkara: Didn't we reserve the keyword `element`?

kylef: Ah, then perhaps this could be `entire` to represent the entire element.

smizell: Just to note, we reserve it for object properties, but not sure about values (here it would be the value of the `path` property). With that said, it may be helpful to not use `element` as a possible value.

Second, do you want to require this value @kylef? I know you have expressed thinking about the Zen of Python, and it would remove a rule for the implementation (i.e. if `path` isn't set, do X, if not, do Y). It doesn't matter to me, so I'm just bringing it up if we want to start thinking about being more explicit with the serialization/deserialization.

smizell (approved):
I like this change. Makes it a lot simpler. I don't feel too strongly about how to handle the `path` value of `element`—whether that should be `element` or `entire`. I think `element` makes the most sense, but I understand if you all think it should be reserved.

I also added a comment on requiring a value for `path` to make it more explicit.

Overall, 👍

pksunkara:
Well, if we are writing down serialisation formats and rules, do we still need to reserve the `element` keyword? (Maybe we do for embedded refract format, but I just wanted to raise the question).

kylef:
Ok so are we good to go with this RFC @pksunkara ?

pksunkara (approved):

pksunkara:
Yup, didn't realise `element` is a value here.

